Issue #32, opened 2012-09-28
Search bar misplaced


ebbes:
Well, I think the search bar is a bit misplaced, especially when looking at the search results pane.
In my humble opinion, it should be located on the right of the bookmarks pane.
See http://www.picfront.org/d/8ORd
Nautilus 3.4 in background, Nemo in foreground.

PR #83, opened 2012-11-03
Layout changes - toolbars and search bars

mtwebster:
- Adds a second toolbar below the main one, to hold the location/path bar.
- Makes search bars slot-dependent.  This means you can have individual search bars visible for each pane or tab you have active, and the views make more sense.  This should address #32.  If I've misconstrued the intent, let me know and I'll adjust.

Screenshot: http://dl.dropbox.com/u/8201170/layoutchanges.png

### DIFF
--- a/src/nemo-location-bar.c
+++ b/src/nemo-location-bar.c
@@ -563,3 +563,9 @@ nemo_location_bar_get_entry (NemoLocationBar *location_bar)
 {
 	return location_bar->details->entry;
 }
+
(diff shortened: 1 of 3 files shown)


zbandhan:
can you design locationbar like this http://4.bp.blogspot.com/-KAYwYxd52tw/Vht1K7dfb_I/AAAAAAAAElg/xAz7wck3EgU/s1600/2.jpg

PR #429, opened 2013-10-29
Advanced search

### LINKED ISSUE #426: "find in files" for Nemo / Integrate gnome-search-tool
It would be nice if Nemo search can support searches like "find in files". 

This can be implemented by a Integrate gnome-seach-tool into nemo. 
The easiest way by launching gnome-seach-tool with an "advanced"  button.

daschuer:
This patch adds an "advanced" button at the left side of the search bar to call gnome-search-tool with the current search. 

This fixes #426

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -44,7 +44,8 @@ Depends: ${shlibs:Depends},
 Recommends: eject,
             librsvg2-common,
             gvfs-backends,
-            nemo-fileroller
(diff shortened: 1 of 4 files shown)


clefebvre:
I don't find this solution very elegant. The button requires mouse interaction, the search isn't initiated, it doesn't work when the search is already performed normally. I don't see the point really.

PR #433, opened 2013-10-31
Make bookmarks sortable

nkoep:
This adds a new public function to the NemoBookmarkList API which allows
sorting bookmarks in ascending order. The function is used in the
"clicked" callback of a new sort button added to the bookmarks window.

Signed-off-by: Niklas Koep <niklas.koep@gmail.com>


mtwebster:
I'm not 100% on the button being labeled 'Ascending,' I think I'd rather it just say 'Sort,' but I'm not going to worry about it for now.

So... a year and 4 months later... (sorry!)

PR #496, opened 2013-12-24
Preference to sort dotfiles and dotfolders first

droidmonkey:
Adds a preference for dotfiles and dotfolders to be sorted first in the displayed list. This is in contrast to the currently hardcoded method of "last sorting" the dotfiles/folders.


clefebvre:
It's a bit subjective whether this is overkill or not. I don't really see the need personally, but then again I don't have a problem with it either.

It should be discussed. The changes are minimal and the rebase is much bigger, so I'm closing this PR for now.

Issue #520, opened 2014-01-27
Feature Request (renaming files by two times clicking on them)


arnaghizadeh:
Hey folks
Long time ago when I was still using windows there was this very nice feature that i'm still missing in Linux world! It is not a big thing but can make your life a lot easier if you know what i'm saying (like that nice new button that opens nemo as root which makes your life easier)!
so here it is I want when we click two times on a file or folder, with a little bit pause between them, it triggers rename of that file or folder! 
For seeing this in action any windows version does work just click on a file, pause a moment and click on it again then you can see that you can rename the file.
Really appreciate if someone does implement it! 
thank you very much!

Chivey101:
Shame! it is such a handy feature in windows, but is differently a feature you love or hate. 
But having more options can only be a good thing.

AjvarXX:
Any "fastpoke" who accidentally triggers rename due to long thinking while doubleclicking could set trigger time window after a whole second... or not to turn it on at all. Why would I use 2 hands for simple rename? Just doubleclick name!

tinyhazeleyes:
Thanks for adding this.  It's so handy.

PR #529, opened 2014-02-03
[WIP] renaming files by two times clicking on them[final version]

AlirezaNaghizadeh:
Adds the feature which was discussed in https://github.com/linuxmint/nemo/issues/520 , here is a video preview http://youtu.be/J9ogU80dwr0

### DIFF
--- a/libnemo-private/nemo-global-preferences.c
+++ b/libnemo-private/nemo-global-preferences.c
@@ -82,6 +82,23 @@ nemo_global_preferences_get_tooltip_flags (void)
     return flags;
 }
 
+gboolean
+nemo_global_preferences_should_load_plugin (const gchar *name, const gchar *key)
+{
+    gchar **disabled_list = g_settings_get_strv (nemo_plugin_preferences, key);
+
+    gboolean ret = TRUE;
+    gint i = 0;
+
+    for (i = 0; i < g_strv_length (disabled_list); i++) {
+        if (g_strcmp0 (disabled_list[i], name) == 0)
+            ret = FALSE;
+    }
+
+    g_strfreev (disabled_list);
+    return ret;
+}
+
 static void
 ignore_view_metadata_cb (GSettings *settings,
                          gchar *key,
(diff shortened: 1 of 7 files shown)


simonzack:
@mtwebster bump.

mtwebster:
Hi, sorry for taking so long to look at this - I'd like to include it for 2.6.  A few issues though:

- I think that if I first click the icon of a file, then click the text, that should count as a second click, and induce renaming.  Right now, if I select the icon, I then have to click the text *twice* to rename.
- It seems like it doesn't matter when I do the second click, it'll always rename - even 5 minutes later - is this intentional?
- Can you rebase this PR against the current nemo master?  (If you don't know how, let me know and I can give you step-by-step)
- Can you implement this in the list view also (same as icon view, only when over actual text)
- Can you do a little bit of code cleanup to make it consistent to itself and the rest of nemo:
  - Use spaces, not tabs (4)
  - Put spaces between all operators and around function parameters - for example, `if (nemo_view_get_foo (view) == view->bar) { }`
  - don't put comments after a `{` - just place the comment before that block of code


Let me know if I'm not making any sense - thanks!

pascal95:
waiting for this important feature for months now
(could be switchable in the config - so no one would be disadvantaged if someone doesn't want it)

AlirezaNaghizadeh:
hey guys, sorry I got a little busy, I will clean the code in may. 

@mtwebster 
1- Agreed, this makes more sense. I will fix it up
2- Yeah it is intentional. It is the same as windows, it is working just fine for years on windows I didn't see any reason to change it. However, making a limited time makes sense too. I'm neutral on this one, if you think it would make more sense I can put some kind of limit on it.  like 100 times the time of double click or something similar. If users wait more than this time they have to click two times again instead of one.
3- yup I will, I will try to find a way myself.
4- hmmm.. doesn't work on list view? didn't know that, lets see what I can do(this may be a little challenging as I don't know what the problem is now).
5- yup, I will keep those in mind.

kevlogan90:
@AlirezaNaghizadeh have you had time to work on this code? This is something I would like to have added to nemo.

AlirezaNaghizadeh:
Hey guys, in one word I'm super busy, there is this visa application process going on and the some papers that I have to finish very soon. The worst part is I don't have any Linux boxes around here right now. If someone can clean the code by all means, please help.

AlirezaNaghizadeh:
I cleaned up the code myself, the final version is pushed with another pull request!
@mtwebster the problem with List View is not related to this code! The code uses "nemo_icon_canvas_item_get_icon_rectangle" and "nemo_icon_canvas_item_get_text_rectangle" as standard functions to get the dimensions of texts and icons! These functions are applicable in Compact View and Icon View but do not work in List View mode. This seems to be a bug in the code which is another problem altogether. I think you guys should fix these two functions to work in List View as well. Because it is a minor issue and unrelated to this topic I have pushed it as it is.

PR #607, opened 2014-05-08
[Branch] Generic Version with conditional Gnome and Unity support

daschuer:
This is a version of current master, which compiles and runs with Ubuntu Trusty. 
It is basically patched with the patches from 
http://www.webupd8.org/2013/10/install-nemo-with-unity-patches-and.html 
So all credits are going to Jacob Zimmermann. 

I have added conditional compiling for Unity, Gnome and Zeitgeist, in a way that it would be possible to maintain this in master branch. 
It should still work with Cinnamon without any issues.


daschuer:
This is now Nemo 2.8.7 for Ubuntu Unity and Gnome.

mtwebster:
Honestly this should be a fork.  I don't care if we support other DE's to this degree, other than as a simple file manager, and I don't want the complexity added to support different libraries, DE's, etc.  I get bug reports that make no sense from the perspective of git Nemo, and it wastes peoples' time trying to figure out what is going on.

The way you're doing this makes it impossible for me to really see the breadth of changes that are made here.  I'd have to clone this branch and diff it against master.  It's bad enough we have debian folks making package name changes that trickle to Ubuntu, I'm not about to wonder why gnome-desktop is behaving differently than cinnamon-desktop, nor do I care, except from the perspective of maintaining the expected behavior of cinnamon-desktop.

I'm not trying to minimize the efforts you've put into Nemo at all (and re-reading this I know I sound a bit Gnome-ish,) but from a practical and technical debt standpoint, we can't afford this generalization.

Things I would be ok with bringing into nemo master:
- supporting file-manager-specific behavior or concerns: progress handling for Unity, etc.. I've wanted this for Cinnamon for some time, I would not mind supporting Unity with a *runtime check.* for support.
- better handling of conditional context-menu items that are DE specific - I don't see why this can't be done altogether in the packaging phase - everything relevant is action-based.  I'd be ok putting another field in the `.nemo_action` files to specify DE (like desktop files), if a dbus check that we have now isn't sufficient.

Things I'm not ok with:
- conditional includes, gnome-desktop, etc
- dbus service name changes
- re-adding background handling

I realize I have no real control over this situation, and I'm absolutely not trying to be an asshole about it, but bug reports all funnel back here, as do reputation and quality concerns.

daschuer:
Ok, this means you are willing to add a striped version from this branch to Nemo master? 

I have tried to use runtime checks as much as possible. 
The conditional gnome includes are "just" required to compile nemo on a plain unity without cinnamon installed. If you target a system where Unity or Gnome runs in parallel with Cinnamon there should not be a big issue. 

How a bout the libzeitgeist and libunity dependency? 

In my recent changes I have removed the original desktop handling that is part of the Ubuntu version of Nautilus. Since Unity stops drawing the desktop once Nemo has placed it's transparent desktop window, we still need a solution to avoid the funny garbage drawing on the desktop. 
I have solved it by a third desktop window that holds the wallpaper and which is shown conditionally on Unity only.

mtwebster:
What makes Unity stop drawing the desktop?  Is there something we can do differently to prevent that?

Or is it just because it's obscuring the desktop?  Hmm

I'll have to think a bit more about the rest, how we're packaging/building.

mtwebster:
Ah just tried it, sorry - without nautilus, there's nothing drawing the desktop backgrounds.  I didn't realize this was still the situation in Unity..  I know it may not be ideal for some, but in this case nautilus should be kept to manage the desktop only, and Nemo disabled from autostarting (via gsettings).  We could add a check at startup that looks at some environment variable to determine whether or not to autostart nemo.

daschuer:
That solution would not work for me, since a click on the desktop folder will fire up a Nautilus window instead of the desired Nemo one. It will also break all the nice Nemo features on the desktop because without patching Nautilus, we cannot stop it from managing the desktop as well. 

Please note that a upstream Nautilus does not draw the desktop background, only the Ubuntu version does.  

A possible solution would be to provide a desktop background drawer application or Improve the mechanism that stops Unity from drawing the background. 
So I have also dgiged though GitHub to find an existing desktop backround drawer. 
What I found was: 
https://github.com/GNOME/gnome-flashback/tree/master/gnome-flashback/libdesktop-background

After all, my conclusion was that it is by far less work and maintenance to adapt the existing Nautilus  patch for drawing the desktop with Unity. It is maintained for free by Ubuntu and only requires a few extra lines in the Nemo code that likely does not hurt us during future changes.

mtwebster:
I was thinking similar, I can write a helper program, 'background-fallback' or something in python, you feed it a schema id (org.gnome.desktop or org.cinnamon.desktop) and it just runs as a simple daemon that uses cinnamon-desktop background api to draw it.  This would keep the work out of nemo and its process, and be simple to maintain (it would be a pretty easy python script).  I can package it with nemo if we want.

daschuer:
Yes that might work. However, it will probably not behave like Unity expects. 
What are the differences between the Gnome and the Cinnamon background API?

mtwebster:
Not much difference, except we've added wrappers for a few common cases that allow use with introspection (avoiding problems with cairo surfaces I think).  As I said, a command line option could specify to use either gnome's or cinnamon's (or anyone's that adheres *more or less* to the background gsettings schema and it would just work.

I'll see if I can hash something together, I've got a bit of code I can use from a python cinnamon-screensaver rewrite I'm working on.  I've been running on xenial lately so it'll be easy to test.

clefebvre:
https://github.com/linuxmint/nemo/pull/1041

PS: Let's tackle the relevant issues separately and in isolation.

PR #696, opened 2014-09-08
Desktop grid

### LINKED ISSUE #108: [Suggestion] Desktop Grid
One of the things that annoys me about Nemo's (and Nautilus') desktop is that there is no proper "grid" of icons, only columns. Both Windows and OS X let you keep a neat grid of icons on the desktop, but with Nautilus/Nemo, your desktop ends up looking very messy if you have a lot of icons, since the rows won't be straight.
If you could add a setting to enable a proper grid with horizontal rows, it would be greatly appreciated.

t-oster:
This pull request is an attempt to fix #108. It is not perfect yet, because it is only a workaround, but I wanted to create a pull request in order to be able to keep track of the changes and have a place for suggestions.

With the current changes, you will get a regular grid, if you disable preview thumbnails AND if you set the dconf-setting "/org/nemo/desktop/text-ellipsis-limit" to 2
TODO:
- [ ] automatically compute grid size from biggest icon
- [ ] maybe reduce preview-icon size to regular icon size?

### DIFF
--- a/libnemo-private/nemo-icon-container.c
+++ b/libnemo-private/nemo-icon-container.c
@@ -103,9 +103,9 @@
 
 /* Desktop layout mode defines */
 #define DESKTOP_PAD_HORIZONTAL 	10
-#define DESKTOP_PAD_VERTICAL 	10
+#define DESKTOP_PAD_VERTICAL 	1
 #define SNAP_SIZE_X 		78
-#define SNAP_SIZE_Y 		20
+#define SNAP_SIZE_Y 		88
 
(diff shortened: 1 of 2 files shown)


mtwebster:
We've discussed this internally, and probably will end up revamping how the desktop is handled entirely - no more pixel positions, grid positions instead.  Closing this, will address the open issues when it's handled.

Mahi:
> We've discussed this internally, and probably will end up revamping how the desktop is handled entirely - no more pixel positions, grid positions instead. 

Any ETA?

PR #782, opened 2014-12-05
Add application-specific theme for Adwaita

City-busz:
This patch adds some desktop style when the Adwaita theme is used. Adwaita is the default theme of GTK+ since version 3.14, so it would be worth if Nemo support it.

Based on this patch for Nautilus:
https://git.gnome.org/browse/nautilus/commit/?id=3fd7e847d531fbbfc4ebba24864bdd0f8b81c750

### DIFF
--- a/src/Adwaita.css
+++ b/src/Adwaita.css
@@ -1,0 +1,26 @@
+.nemo-canvas-item {
+    border-radius: 5px;
+}
+
+.nemo-desktop.nemo-canvas-item {
+    color: @theme_selected_fg_color;
+    text-shadow: 1px 1px black;
+}
+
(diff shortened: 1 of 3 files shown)


clefebvre:
This is wrong.

The theme has no idea what background is being used, and therefore how to render text that would be legible.

Experience showed the best solution for this was to show white text on top of a black shadow. We should use the mint-X style for all themes here and retire themeing for the text rendering.

We'd then end up with the same rendering everywhere, no matter what theme is selected.

mtwebster:
https://github.com/linuxmint/nemo/commit/e59cae7bd231bac01806458b377cd8c388503751

PR #783, opened 2014-12-05
Rename 'Files' app name to avoid having the same as nautilus

City-busz:
Currently both Nemo and Nautilus use the same name in the desktop launcher ('Files'), which may confuse users. I think it would better to rename the desktop file of Nemo to 'Nemo'.


clefebvre:
In most cases people only run one DE. Names should be generic as much as possible, so both projects are using the right name imo. We can consider adding an OnlyShowIn though, or making the menu smarter so it shows more info when two items have the same name.

PR #801, opened 2014-12-21
raise thumbnail size limit

leigh123linux:
Raise thumbnail size limit to account for larger rip sizes of HD media.


dalcde:
Can we make a "custom" or "unbounded" entry?

PR #823, opened 2015-01-24
Only include the domain if the method is SMB (bugfix)

Nemilar:
Fixes a bug where if the user first tries the SMB method, enters a domain, and then tries another method (e.g, ssh), the code would incorrectly prepend the domain to the username.


Nemilar:
Judging from the dialog box, it seems that domain is only used for SMB (the field is not shown for the pother protocols).  But, I don't know much about the other protocols, so I am not 100% sure that is correct.

mhines-datto:
:+1:

PR #833, opened 2015-02-03
add a simple transition when toggling the location bar/pathbar

JosephMcc:


daschuer:
After this merge I get the message: 
when starting with location entry and switch to pathbar.

PR #838, opened 2015-02-16
[WIP] Fix gtk_widget_draw assertion

daschuer:


daschuer:
This patch does not fix the issue. 
The assertion appears, if you start with the location entry and switch to the path bar. 
if you start with the path bar and switch to the location entry the issue does not happen. 
All following toggle actions are working without the issue. 
It is a regression from: 
https://github.com/linuxmint/nemo/pull/833

daschuer:
I am still seeing this any idea how to fix?

clefebvre:
It happens also if you take the focus away from the location entry, and then CTRL+L again.

PR #840, opened 2015-02-17
Add Configurable Context Menus

RavetcoFX:
+Configurable items in the context menu
(Uses binary flags to keep the gsettings down to a minimum)
+Better logic when to show some context items instead of making them insensitive (i.e Duplicate and Make Link)

TODO:
-Implement a nice looking GUI that emulates the appearance of the menu


RavetcoFX:
First bit of work on the GUI:

screenshot from 2015-02-25 00 40 36

PR #860, opened 2015-03-10
[WIP] Make Nemo search files automatically when the contents change and display the results at the same time

Jason-King1992:
I modified something so that the nemo search bar can search files automatically when the contents change and display search results incrementally at the same time.
And I fixed the bug that the nemo search bar can not remember and restore the focus when first input string.Now, the nemo search bar can continue to receive input when the contents change.


clefebvre:
needs rebasing.

daschuer:
The same feature will be land when we continue to merge the Nautilus search. 
Please have a look if you can use some of the rebased commits in https://github.com/linuxmint/nemo/pull/1011
This will avoid conflicts if we try to merge other Nautilus features.

PR #867, opened 2015-03-30
Nemo Toolbar: Change the rest of the icons to symbolic

RavetcoFX:
screenshot from 2015-04-15 10 17 35


lots0logs:
@mtwebster What do you think of this? I came here to see if there were any requests for this to be done and planned on asking you how I could help get it done. I am pleasantly surprised to see someone beat me to it! :smile:

PR #884, opened 2015-05-04
[3.0] Add actions for extract and compress

RavetcoFX:
This would replace the NemoFileRoller action, allowing it to be enabled/disabled without restart.


GLolol:
This would require a hard dependency on File Roller, which I'm not sure is desirable...

GLolol:
I'm wondering if it would be simple to allow other archivers to integrate with Nemo, similar to Thunar and PCManFM.

RavetcoFX:
@GLolol this should be very easy to do with actions as long as other archive managers support command arguments

GLolol:
How? Just adding a `nemo_action.in` in the source?

Fantu:
This will fully integrate/replace nemo-fileroller?

RavetcoFX:
@Fantu yes

Fantu:
This PR seems need rebase.
Need also a change in debian/control: nemo-fileroller->file-roller
File-roller have too many unuseful dependencies, for example libnautilus-extension, nautilus-data and probably other
Probably is good try to find an alternative that don't install nautilus or other unuseful gnome things.

Fantu:
In debian we contacted file-roller maintainers about the nautilus and other gnome things taken by it as deps but unused in major cases of other filemanager/DE and they removed nautilus-data from deps (in 3.16.2-2) that is the mainly problem.
Now seems good remain with file-roller without have the problem of nautilus and other unused things unistallable.

PR #913, opened 2015-06-19
Quickrename

AlirezaNaghizadeh:


AjvarXX:
I am waiting for this feature... Can't wait!

Although I would add some feedback. Would be cool if there was an option to set time window for rename. Starting time after first click and finish time, where 0 means default and any number means milliseconds.
For example 200-1000 means any time you click between 200-1000ms after first click it renames. This feature can be added later though.

merlinux2004:
Thank you, AlirezaNaghizadeh

great work!

AlirezaNaghizadeh:
thanks guys, I hope @mtwebster merges this feature quickly! last time it took them one year to respond and I got a lot of PMs requesting to update the code. It seems a lot of people want this feature and I don't have time to update the code every now and then.

AjvarXX:
Can I easily update/merge my existing Mint installation with this? It's better than wait when those snobs finally accept force of demand because looks like they will never admit that they are wrong and people need this option.

JosephMcc:
@AjvarXX Wow, thanks for you're super helpful and productive comment!

mtwebster:
I didn't merge it yet because it's not implemented in list view - 

See my feedback on the original PR: https://github.com/linuxmint/nemo/issues/529....
- Can you implement this in the list view also (same as icon view, only when over actual text)

I do plan on merging it, hopefully for the next version - I agree it's a useful optional feature, particularly for those coming from Windows - but I'm not going to do it until I have time to complete the implementation in all views, or the author does.

AlirezaNaghizadeh:
Hey, as I mentioned before, the problem @mtwebster has mentioned, seems to be how nemo recognizes the dimensions of icons not how this code should work. Sorry that I don't really have time to work on this project anymore.
Best

AjvarXX:
If the problem is in strict area of  2nd click (name area) then for beginning it could be just taken away so that rename happens on 2nd click whenerver it's clicked. And increase the minimum time after first click to recognise it a Rename click.

Maybe I sound like redneck who wants everything for free but this is to become the case where this idea drops forever and nobody will ever make it happen. Just it.

clefebvre:
Ok, this is incomplete... (it's missing in the list view basically)

.. but it works quite well.
.. and it's disabled by default.
.. and it won't harm the experience for people who don't care about it..
.. and people who do care about it probably want it even if it's not complete just yet.

So with all that in mind, let's merge it.

AjvarXX:
YAAAAAAAAAAAAAY!!!!
Thank you! Even though you might not see it just yet, it's a big step ahead for Linux Mint.
It's possible that after it appears in new Mint some developer decide to invest their free time to make this feature complete. And it brings lost usability for those who jumped on Linux recently.
BIG TNX!

merlinux2004:
Finally, my girlfriend and family won't have any excuse to switch to Linux.
Thank you.

simonzack:
This is truly a giant step for desktop linux. Big congrats on lethargic87 and nemo for being the first file ever manager on linux to implement sane file renaming, you guys rock!

lethargic87:
I received the romeo update today! There are three concerns:
1- Now that I have seen some comments from Reddit maybe we should rename it to "click rename" instead of "quick rename"! It makes more sense!
2- I don't know what happened but it is not working as it should! The second click should only be in the text area but now even if I click two times on the image part, it gets renamed! I don't know if it was me who made a mistake or something in nemo has changed?
3- Again it is not working as it should like if you deselect an icon and click on it again, it goes into renaming mode.
The last two problems were not present in my working example, I don't know maybe somewhere I made a mistake but they should be fixable easily.

PR #962, opened 2015-09-01
replace margin-left & margin-right

leigh123linux:


JosephMcc:
Margin_start and margin_end weren't added until gtk3.12 so this is incompatible with Mint17.

PR #974, opened 2015-09-24
Re-add revealer animation on search bar

RavetcoFX:


clefebvre:
Why was it removed? And why do you think it can be readded?

RavetcoFX:
This commit https://github.com/linuxmint/nemo/commit/4d50cffdcafc61d58ddf73437ae69bf0f408d523 removed it, and it was a nice visual effect

daschuer:
The revealer will be part of the final search solution: 
https://github.com/daschuer/nemo/blob/nemo_312_merge_nemo_base/src/nemo-window-slot.c#L342

If I find time, I am going to do small PR towards the this final solution.
However these small iterations are a lot of work and is not much fun doing the work twice facing a fully working and merge-able branch. 

However, this PR can fix the issue until we merge the final solution. 

@RavetcoFX: Would you mind to replace the four spaces indent by tabs, to fit to the rest of the file. 
Thank you.

PR #988, opened 2015-10-19
Add control to context menu to toggle simple/complex menu

mtwebster:
screenshot-monitor-2015-10-19-081146

not ready... just concept - need to:

- always show all entries in pulldown menus
- only show widget in selection menu (not background click)
- small GC issue

### DIFF
--- a/data/icons/Makefile.am
+++ b/data/icons/Makefile.am
@@ -18,6 +18,8 @@ public_icons = \
 	hicolor_actions_scalable_sidebar-show-symbolic.svg \
 	hicolor_actions_scalable_sidebar-tree-symbolic.svg \
 	hicolor_actions_scalable_sidebar-places-symbolic.svg \
+	hicolor_actions_scalable_expand-menu-symbolic.svg \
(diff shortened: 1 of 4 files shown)


clefebvre:
Nice! 

Couple of comments re-visuals:

- Have a different icon for collapse/expand, maybe just a little + or - sign
- Add a tooltip to it if possible

mtwebster:
Works better now, most issues fixed, just need a tooltip and fix the menu position when re-showing back to the original popup location.  Maybe better icons, not crazy about the gtk-remove icon.

mtwebster:
plus
minus

clefebvre:
Looks nice, icons are way too big though.

clefebvre:
This can go in, it works very well. If we merge it, it has to be before the Mint RC and it needs the following:

- The icons should be smaller.
- Tooltips should be more explicit "Show more options", "Show less options".

And then, not needed, but would be nice at some stage:

- The icons should be within a button which border is revealed on hover. Similar to the color buttons in the folder color extension. This would make it look better, and also allow the clickable zone to be as big as it is now, while making the icon smaller.

PR #989, opened 2015-10-22
Two suggestions

A-Shahbazi:
I have implemented two suggestions for nemo file manager:
* I have used elementary freya os for a long time now and the pantheon-files(the desktop's default file manager) is really nice in terms of design. One of the things which makes a big difference in my opinion is that they use 64px icons instead of 48px for their default view. I changed default nemo icon size from 48 to 64 here are two screenshots:

48px icons:
screenshot from 2015-10-20 12 22 58

64px icons:
screenshot from 2015-10-20 12 35 39

* I thought "File System" partition should go under Devices section like other file managers.


clefebvre:
Thanks, but please do separate different work items into different pull requests.

A-Shahbazi:
first suggestion PR: linuxmint/nemo/pull/990

PR #990, opened 2015-10-23
[3.0] use larger icon size for standard view(64)

A-Shahbazi:
I have used elementary freya os for a long time now and the pantheon-files(the desktop's default file manager) is really nice in terms of design. One of the things which makes a big difference in my opinion is that they use 64px icons instead of 48px for their default view. I changed default nemo icon size from 48 to 64 here are two screenshots:

48px icons:
screenshot from 2015-10-20 12 22 58

64px icons:
screenshot from 2015-10-20 12 35 39


Magissia:
Wouldn't it be smarter to set default size based on screen DPI ?

JosephMcc:
I like it the way it is currently. I've tried the newer nautilus with the large icons and I'm not a fan. I don't feel the need to maximize my nemo window to see more the 6 or 8 items.

A-Shahbazi:
It's not about showing less items. It's about the blank spaces between items. In both of attached images I got 7 columns of items. Actually, the waste space(blank space between items) is being used and it provides a better look.

PR #1023, opened 2015-11-26
improve context menu toggle:

mtwebster:
- change icon style when hovering over it
- change menu text and right-justify "Show more/less items" (instead of a tooltip)
- Restore icon and text updates to the menu item (never changes previously)

togg1
togg2

### DIFF
--- a/libnemo-private/nemo-context-menu-menu-item.c
+++ b/libnemo-private/nemo-context-menu-menu-item.c
@@ -1,0 +1,422 @@
+/* GTK - The GIMP Toolkit
+ * Copyright (C) 1995-1997 Peter Mattis, Spencer Kimball and Josh MacDonald
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ */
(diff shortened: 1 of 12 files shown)


daschuer:
Thank you for the PR. This improves the situation slightly. 

IMHO the oiginal issue remains, since the new hovered view only appears when you are already tricked.
Thhe "+" is still highlighted when you select the row. 
It is not unambiguous, that the "show more actions" is related to the menu and not to the file in the open context. 

I have played with gimp a bit and that was my result: 

Current view: 
original

Draft: 
draft

libnemo-private/nemo-context-menu-menu-item.c
```diff
@@ -0,0 +1,431 @@
...
+                         G_IMPLEMENT_INTERFACE (GTK_TYPE_ACTIVATABLE,
+                                                nemo_context_menu_menu_item_activatable_interface_init));
+
+static void
+update_toggle_state (NemoContextMenuMenuItem *item, gboolean from_event)
+{
+    gboolean complex_mode = g_settings_get_boolean (nemo_preferences, NEMO_PREFERENCES_CONTEXT_MENUS_SHOW_ALL_ACTIONS);
+
+    gtk_label_set_text (GTK_LABEL (item->toggle_label_widget),
+                        complex_mode ? _("Show less actions") : _("Show more actions"));
```
daschuer: It is not clear that the context is the menu and not the open action
How about: 
"Expand menu" "Collapse menu" 
or 
"Simple menu" "Advanced menu"

daschuer:
This PR has 2 Space indentation in some functions. I think you use normally 4 pace in your changes. 
However all other indentations are a single tab.

libnemo-private/nemo-context-menu-menu-item.c
mtwebster: I didn't want to upset translations at this point (I'll probably try to get this in 2.8.x still) - but you're right, it could be worded better I think

mtwebster:
I see that - I'll fix indentation.  Copy/paste laziness.

mtwebster:
merged here: https://github.com/linuxmint/nemo/commit/fb97b62aad23bb44820713c8084785bbc147bfd7

Issue #614, opened 2014-05-18
Incorrect URIs created for filenames with spaces for custom bulk rename command


artfulrobot:
I'm writing a little app to handle nemo's bulk rename operation. Unfortunately on testing I'm finding that the URIs passed to the command as arguments are not valid for paths with spaces.

e.g. this test script: 

```
#!/bin/bash
echo "$@" >/tmp/test.log
```

use this for the rename script, select two files inside a directory with a space, i.e.

```
/home/foo/test/a b/a
/home/foo/test/a b/b
```

and you get something like this in the log file:

```
file:///home/foo/test/a0b/a file:///home/foo/test/a0b/c
```

I would have expected `%20` to replace the spaces, instead I'm getting `0`. So I would have expected:

```
file:///home/foo/test/a%20b/a file:///home/foo/test/a%20b/c
```

This appears to be a bug, and it makes my rename script useless :-(

PS. I should say that I'm using nemo 2.2.0 from http://ppa.launchpad.net/noobslab/mint/ubuntu/ on Ubuntu Gnome 14.04.

gritty-gitty:
I can confirm that *nemo* (here v.2.4.5 in Rebecca) doesn’t pass a correct URI string to the bulk renaming command if the filename contains white or UTF-8 characters. It does paste any filename correctly to the clipboard tough:
`File name:      /home/…/d’après moi`
`To clipboard:   file:///home/…/d%E2%80%99apr%C3%A8s%20moi`
`To renamer:     file:///home/…/d209apr38s0moi`

Hence the escape character and the first UTF byte always go missing.

gritty-gitty:
Maybe a hint: the configuration key *bulk-rename-tool* in the scheme *org.nemo.preferences* and its type are both messed up. The first reads *b'nemo_bulk_rename.sh'* and the second *ay*, and neither of them can be set from *dconf*.

LeInterwebsFTW:
The problem is that g_app_info_create_from_commandline unquotes based on the Desktop Entry Exec key and eats the percent signs. To prevent this, all percent signs need to be doubled. A quick fix:
```
--- a/src/nemo-view.c
+++ b/src/nemo-view.c
@@ -7388,6 +7388,15 @@ invoke_external_bulk_rename_utility (NemoView *view,
                g_free (quoted_parameter);
        }
 
+       // Escape percents
+       for (int i = 0; i < strlen(cmd->str); i++) {
+               if (cmd->str[i] == '%') {
[...]

PR #1091, opened 2016-02-11
Escape Percents in bulk rename

ChrisJAllan:
Fixes issue #614

`g_app_info_create_from_commandline` eats the percent signs in file names with spaces. This doubles all percent signs.

This works better than `g_uri_unescape_string` as it still works on file-names with percent signs in them.


brownsr:
shouldn't  the character insert be on cmd->str rather than on cmd ?

Issue #802, opened 2014-12-21
Reducing the horizontal width/spacing between files to utilize more space


enriquezmark36:
The spacing between the file and folders in a regular window, which is supposedly more than enough to fit an icon of a file or folder, is quite large and might be a bit annoying if one have to look to a lots of files with thumbnails.**(Illus.1)** Zooming down  doesn't help since the gap is still quite large by comparison plus the icons shrink down too.
The current one looks neat but wouldn't it be better if it can be reduced or changed via the code or the settings?

**Illus.1**
![screenshot from 2014-12-21 22 05 37](https://cloud.githubusercontent.com/assets/7193373/5518352/20806072-895e-11e4-924f-431de1432494.png)

joe1970:
+1, I agree, this behaviorto leave that much space unused is strange. There is however the compact layout for the icon view. It can be toggled in "arrange Icons" in the context menue. It seems to pack icons next to each other as long as there is still space in a line left. The result is a nicely packed view, but icons are not arranged in a grid any more.

ManIVIctorious:
@joe1970 i know about this one, but that destroys the alignment of the following row. I'd like to see a better alignment or an option to change the spacing manually (doesn't even have to be exposed, a gsettings call would be enough for me)

enriquezmark36:
@ManIVIctorious, @joe1970, After studying the code for about 3 days there seem to be a way to reduce the width and make use of the space like this: 
![nemosend](https://cloud.githubusercontent.com/assets/7193373/12535061/01c59f98-c2b1-11e5-8edd-fdd255d783bd.png)
It's a work-around and it's summarized to this patch: http://pastebin.com/raw/hAVjxJYr, and since I don't really know C, I copy-pasted a lot from the source. 
The patch will create 3 options and can be found at org->nemo->icon-view: icon-width, custom-icon-width and custom-max-text-width.

JosephMcc:
I would be okay with a bit less spacing as long as it wasn't too cramped but having so many options seems a bit over the top. If it's going to be changed, why not just change it and call it good?

enriquezmark36:
Yeah, you're right. Too much options are confusing; it's also the reason I'm having a difficulty on plasma 5. I originally plan to add only 1 option as there was supposedly no "custom" options since its a bit complicated but will enable someone to manually tune the width even further.

enriquezmark36:
I created a fork where the patch is applied, it's in here: https://github.com/enriquezmark36/nemo.
[...]

PR #1095, opened 2016-02-16
Add the ability to change width between files and maximize the space

enriquezmark36:
Well, the details are described in #802 but this is the simpler version of the patch.


enriquezmark36:
I'm closing this for more revisions, something is still missing in this.

Issue #1009, opened 2015-11-14
New style tabs for "Properties" window.


Vahan86:
The file/folder "Properties" window uses old style tabs, whereas "File Management Preferences" window uses new style tabs which are more elegant. It would be nice to have the new style tabs for "Properties" window as well.

Thank you.

Vahan86:
@tomasbruna, thank you very much for implementing the requested feature.

@mtwebster, can we close this Issue?

JosephMcc:
Done

PR #1102, opened 2016-02-22
New style tabs for "Properties" window

tomasbruna:
Use a stack instead of notebook

Implements feature request #1009


ChrisJAllan:
The top bar has a boarder around it as seen [here](https://cloud.githubusercontent.com/assets/9275775/13281383/62e3f8b6-da98-11e5-8595-8a1b0657c7a6.png). The gradient should go to the edges of the window.

tomasbruna:
Thanks, I did not notice that. The border should be gone now.

PR #1107, opened 2016-02-27
nemo-trash-bar: check for null before disconnecting signal

itzexor:
Fixes a few warnings and occasional crash when changing view from trash


mtwebster:
I don't see where this would avoid anything - `FALSE == 0` here.

I think this will do fine though: https://github.com/linuxmint/nemo/commit/e00711a08d6dd74d34a89f906686aa9b5d937763

PR #1115, opened 2016-03-09
GTK 3.19-fix partial desktop rendering

lukefromdc:
Fix desktop rendering on 1/4 screen in GTK 3.19 from current GIT master caused by https://git.gnome.org/browse/gtk+/commit/?id=cdc580463e409a9b7e5f1480afa68d875ff3ff65

If this causes issues in older GTK versions it can be put behind a GTK 3.18 and earlier only build time switch like my Caja work. Comments in Caja's code imply this should not be needed, but Caja has to support GTK2 and Nemo does not.


lukefromdc:
Closing this in favor of 
https://github.com/linuxmint/nemo/pull/1121
Which fixes this the same way as 
https://github.com/GNOME/nautilus/commit/cf79068dc70c87e963e217e2e8ae149ee5ed1872
in Nautilus

PR #1121, opened 2016-03-16
 GTK 3.20-fix wrong desktop size

lukefromdc:
Reset default size of desktop window to -1 in GTK 3.19/20 only. The combination of a non-resizable
window with a default size is what causes the desktop to open wrong size.

based on GNOME/nautilus@cf79068

"desktop-window: avoid wrong size"

"gtk+ recently changed and unresizable windows now take into account
the default size set. We set the default size for normal resizable
windows in nautilus."

"However the desktop window is a no resizable window. What happens now
when setting a default size is that that size is used instead of the
widget request size, since window sizing now takes precedence over
widget sizing."

"To fix it, set the default size to -1 to override what the nautilus
window does by default, so we can avoid this misbehaviour."


lukefromdc:
GTK 3.19 commit 
https://git.gnome.org/browse/gtk+/commit/?id=3e3d29f7b73381dd328e5213e0dca04d3e4f8048
did not fix this, as all it does is special case windows with both a size request and a default size, which helps only current Nautilus.  Thus, the fix Nautilus devs wrote and discarded in favor of GTK's upstrream fix for them is still needed for Nemo.

PR #1131, opened 2016-03-30
GTK 3.20 :: Add style classes to nemo's window, desktop window, and sidebar.

lots0logs:
@mtwebster @clefebvre This assigns style classes to `NemoWindow`, `NemoDesktopWindow`, & `NemoPlacesSidebar` so that Nemo can be properly styled by gtk themes while running GTK 3.20 and later versions. I took the opportunity to do a bit of code formatting cleanup around the three areas where I made the change for the gtk style classes. I made separate commits so its easier to tell which changes are really changes and which are just code cleanup.

Let me know if this looks good or if any changes are needed.  Cheers!

PR #1141, opened 2016-04-13
GTK 3.20 :: Add style classes to nemo's window, desktop window, and sidebar

lots0logs:
@clefebvre @mtwebster Replaces #1131

### DIFF
--- a/src/nemo-desktop-window.c
+++ b/src/nemo-desktop-window.c
@@ -165,6 +165,9 @@ nemo_desktop_window_init (NemoDesktopWindow *window)
 
 	g_object_set_data (G_OBJECT (window), "is_desktop_window", 
 			   GINT_TO_POINTER (1));
+
+	/* Make it easier for themes authors to style the desktop window separately */
+	gtk_style_context_add_class (gtk_widget_get_style_context (GTK_WIDGET (window)), "nemo-desktop-window");
 }
 
 static gint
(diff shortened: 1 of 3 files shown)


lots0logs:
@clefebvre @mtwebster What's the status on this?

JosephMcc:
I can't answer your question about the status. I do know the addition of the "places-sidebar" style class in unneeded.

lots0logs:
@JosephMcc That's your opinion. IMO as someone who currently co-maintains a popular gtk theme, its needed. Also, there's is absolutely zero cost (downside/cons) associated with the inclusion of the style class so why not make the lives of theme developer's a little bit easier? Just my two cents.

JosephMcc:
I'm familiar with gtk theming. How would it add anything different than using ".nemo-window .sidebar" besides having another style class doing the same thing?

lots0logs:
`.sidebar` is not the same node as `.places-sidebar`. Obviously there are other ways to target that node. I was simply trying to point out that what you said was an opinion even though you were very *matter-of-fact* in the way that you said it.

JosephMcc:
I understand and admit I'm not familiar with all the theming changes in 3.20. If you want a class to specifically target the nemo sidebar I would prefer something like `nemo-sidebar` since it is specific and won't show up being used by something else. I'm not a fan of the idea of adding something with such a general name when it's already possible to target it with your addition of `nemo-window`.

clefebvre:
Hi guys.

@lots0logs can you prefix the sidebar class with "nemo" to make it specific or remove it altogether?

Other than that this is ready to merge.

lots0logs:
@clefebvre Done :smile:

clefebvre:
Thanks :)

PR #1147, opened 2016-04-20
Fix compile error and warnings since 4b02ab4

daschuer:


clefebvre:
Already fixed by https://github.com/linuxmint/nemo/commit/6623a66473b90041d2aa56299bdadd11abe592dc

daschuer:
No, sorry, the warning is still there.
We need:  
```
gtk_style_context_add_class (gtk_widget_get_style_context (GTK_WIDGET (window)), "nemo-window");
```

including the cast to GTK_WIDGET

clefebvre:
Ah ok, I didn't spot that. It's missing elsewhere as well then.. a sidebar class was added too.

PR #1149, opened 2016-04-21
Don't draw the desktop on a cloned monitor

daschuer:
If you clone (mirror) the two monitors, both desktops of both monitors are drawn on both monitors. 
This is no harm if the empty monitor is at index 1. If the primary monitor is on 1, the empty desktop is on top of the desktop icons and catches all mouse actions.

### DIFF
--- a/libnemo-private/nemo-desktop-utils.c
+++ b/libnemo-private/nemo-desktop-utils.c
@@ -53,3 +53,28 @@ nemo_desktop_utils_get_monitor_for_widget (GtkWidget *widget)
 
     return monitor;
 }
+
+gboolean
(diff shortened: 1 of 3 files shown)


clefebvre:
@daschuer I can't reproduce the issue. Can you describe how you do that? I've got two monitors here so I should be able to see it.

daschuer:
I do not have a explicit receipt for it. It happens if you play with the "Mirror" and "Main Monitor" settings in 
cinnamon-settings. 

The issue is pretty obvious, if you place a breakpoint or g_print() in the desktop windows creation code. 
You will see that even with mirror enabled, both desktop windows are created.

Issue #1145, opened 2016-04-19
Nemo crashes when adding a column to search result list


jf99:
Steps to reproduce:
- Start a search in any folder
- Right-click the column-title bar
- In the context menu add another column e.g. "Date modified"
-> segmentation fault in nemo

This happens on Arch with Nemo 2.8.7-4.

Related to that: When you remove a column via the same context menu, the column disappears but when you open the context menu again, it is still selected (Look at the attached screenshot where I removed "Size" but it's still there). Removing another column makes nemo crash.
![nemo-search](https://cloud.githubusercontent.com/assets/7962377/14638076/d4299062-0635-11e6-8be0-43174f7ef177.png)

f2404:
Works for me in 3.0.0.

jf99:
Nope, still an issue in 3.0.0 for me. However it does not happen on another machine (with manjaro). So it's apparently system specific. Can I provide any more useful information?

f2404:
@jf99 Could you post the crash info? It can be found in `journalctl` if your system is using systemd.

jf99:
```
systemd-coredump[4143]: Process 794 (nemo) of user 1000 dumped core.

Stack trace of thread 794:
#0  0x00007fd4eef488e0 n/a (libgtk-3.so.0)
#1  0x00007fd4eef16a4b n/a (libgtk-3.so.0)
#2  0x00007fd4eef1e0b0 gtk_tree_view_insert_column (libgtk-3.so.0)
#3  0x0000000000456a57 n/a (nemo)
#4  0x0000000000457c2a n/a (nemo)
#5  0x00007fd4ed350fa5 g_closure_invoke (libgobject-2.0.so.0)
#6  0x00007fd4ed362ff1 n/a (libgobject-2.0.so.0)
#7  0x00007fd4ed36bd8c g_signal_emit_valist (libgobject-2.0.so.0)
#8  0x00007fd4ed36c0bf g_signal_emit (libgobject-2.0.so.0)
#9  0x00007fd4eed1f1c1 n/a (libgtk-3.so.0)
#10 0x00007fd4ed350fa5 g_closure_invoke (libgobject-2.0.so.0)
#11 0x00007fd4ed362b2c n/a (libgobject-2.0.so.0)
#12 0x00007fd4ed36bd8c g_signal_emit_valist (libgobject-2.0.so.0)
#13 0x00007fd4ed36c0bf g_signal_emit (libgobject-2.0.so.0)
[...]

PR #1159, opened 2016-04-26
Fix segfault on changing the number of colums when in list view

### LINKED ISSUE #1156: Nemo segfault on specifics folders (gtk issue)
Nemo segfault on specifics folders.
It happens on some specifics folders everytime (large folders maybe? or with some specifics files in there ?) when it doesn't happens on others folders.

Here the segfault message when I open one of this specifics folders :

> (nemo:5778): Gtk-CRITICAL **: gtk_widget_set_parent_window: assertion 'GTK_IS_WIDGET (widget)' failed
> (nemo:5778): Gtk-CRITICAL **: gtk_widget_set_parent: assertion 'GTK_IS_WIDGET (widget)' failed
> Segmentation fault (core dumped)

Seems to be appearing since nemo 2.8.7 or the gtk3 3.20.3 update.
And it's still reproducible in nemo 3.0.0

Archlinux / 64bits / Cinnamon 3.0.1 / Nemo 3.0.0 / gtk3 3.20.3

f2404:
Fixes #1156 

So I figured out that `gtk_tree_view_column_get_button (l->data)` returned NULL after a column had been removed from the tree at line 1988 - this caused the segfault.
I'm suggesting to hide and show the columns instead of removing and adding them. This fixes the segfault for me and still allows to have different number of columns for different locations (e.g., Home and Trash as described in the issue).

### DIFF
--- a/src/nemo-list-view.c
+++ b/src/nemo-list-view.c
@@ -1985,7 +1985,7 @@ apply_columns_settings (NemoListView *list_view,
 	old_view_columns = gtk_tree_view_get_columns (list_view->details->tree_view);
 	for (l = old_view_columns; l != NULL; l = l->next) {
 		if (g_list_find (view_columns, l->data) == NULL) {
-			gtk_tree_view_remove_column (list_view->details->tree_view, l->data);
+			gtk_tree_view_column_set_visible (l->data, FALSE);
 		}
 	}
 	g_list_free (old_view_columns);
@@ -2001,6 +2001,7 @@ apply_columns_settings (NemoListView *list_view,
                              G_CALLBACK (column_header_clicked),
                              list_view);
         }
+        gtk_tree_view_column_set_visible (l->data, TRUE);
 	}
     g_list_free (old_view_columns);
 


f2404:
@clefebvre Does not the solutions look right to you?

itzexor:
More info on the issue: it only occurs with GTK > 3.18. I believe it is this GTK change specifically that causes this behavior: https://github.com/GNOME/gtk/commit/20ce0588b181f6049af4f5fdc7a3bf0693c91842#diff-f9b0a335e31535d864dc889ab5756acbL1409

This proposed code was reverted in nemo in 2013, though I don't yet know why that is. https://github.com/linuxmint/nemo/commit/8e6e57239314329eaf9b9efa623173408fd23ca1 

I asked mtwebster about it but he hasn't been around since I asked so still waiting to figure that out. If it is re-revertable, then i think we should just do that.

itzexor:
By the way thanks for tracking down this issue and confirming that fixes it, I had some suspicions but I hadn't fully tracked it down yet.

f2404:
Ah, so it's not that Nautilus has fixed it, but Nemo has made changes to the original code. Yes, would be interesting to know why.

itzexor:
I had some time to quickly test your patch and it does fix this issue for me as well.

f2404:
May also fix #1145

PR #1161, opened 2016-04-27
Fix usage of uninitialized value

f2404:
scale will contain garbage if item is NULL.


clefebvre:
It's not used when pixbuf is NULL though.

f2404:
Yeah... I will recall the change.

PR #1163, opened 2016-04-28
Port python scripts to Python 3

mcyprian:
All scripts now Python 2-3 compatible:
- /data/extract_action_strings
- /data/merge_action_strings
- /files/usr/share/nemo/actions/myaction.py

### DIFF
--- a/data/extract_action_strings
+++ b/data/extract_action_strings
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/python3
 
 import os
 import glob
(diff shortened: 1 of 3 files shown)


Fred-Barclay:
Do you think it would be better to use "#!/usr/bin/env python3" ?

clefebvre:
@Fred-Barclay No, I don't think so. If a user was to install an alternative version of python, he'd probably want to run it with his own scripts, but I don't think it should impact system tools and the operating system.

PR #1178, opened 2016-05-18
prefs: Don't try to hardcode a minimum size for the preferences dialog

JosephMcc:
This improves a bit on https://github.com/linuxmint/nemo/commit/aa19ff3ebdb29781a7a3cbd701c589d67ac707c5

Since a large number of user settings such as translations, font-scaling, etc. can affect this, don't attempt to guess a minimum size. Instead get the size wanted by the stack switcher plus a small value to compensate for padding on the toolbar and use that to determine the minimum size instead.

### DIFF
--- a/src/nemo-file-management-properties.c
+++ b/src/nemo-file-management-properties.c
@@ -107,6 +107,8 @@
 
 #define W(s) (gtk_builder_get_object (builder, s))
 
+#define TOOLBAR_PADDING 20
+
 static const char * const default_view_values[] = {
 	"icon-view",
 	"list-view",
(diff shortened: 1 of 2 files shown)


src/nemo-file-management-properties.c
```diff
@@ -765,6 +767,7 @@ static  void
 nemo_file_management_properties_dialog_setup (GtkBuilder *builder, GtkWindow *window)
 {
 	GtkWidget *dialog;
+    GtkWidget *stack_switcher;
```
f2404: The indentation isn't aligned to other lines.

JosephMcc: I know. We normally use a convention of 4 spaces for indentation in new code. It displays properly in the editor but not here.

PR #1184, opened 2016-05-21
fix https://bugzilla.gnome.org/show_bug.cgi?id=668674

leigh123linux:
https://bugzilla.gnome.org/show_bug.cgi?id=668674


clefebvre:
Couldn't reproduce.

PR #1185, opened 2016-05-22
python2/3 compatibility fixes

joequant:
This adds code to make sure that the python3 fixes to cinnamon also work
with python2.  This is important since cinnamon could be building on a
python2 only system.

### DIFF
--- a/data/extract_action_strings
+++ b/data/extract_action_strings
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/python
 
 import os
 import glob
--- a/data/merge_action_strings
+++ b/data/merge_action_strings
@@ -1,10 +1,16 @@
-#!/usr/bin/python3
+#!/usr/bin/python
 
 import os
 import glob
 import polib
 import codecs
 from gi.repository import GLib
+import sys
+# Python2 compatibility
+if sys.version_info <= (3, 0):
+    reload(sys)
+    sys.setdefaultencoding('utf-8')
+
 
 GROUP = "Nemo Action"
 


leigh123linux:
This change would make it python 2 only on fedora 

```
$ /usr/bin/python --version
Python 2.7.11
```

f2404:
This only means that Fedora wants the scripts to be run by python2 by default, right?

leigh123linux:
@f2404 I would prefer pyhton3 for fedora

f2404:
@leigh123linux I'm not using Fedora but maybe this could help? https://linuxconfig.org/how-to-switch-between-python-versions-on-fedora-linux

leigh123linux:
@f2404 None of that will help as I'm not building cinnamon locally, I use the fedora build server (koji) for builds.

https://koji.fedoraproject.org/koji/buildinfo?buildID=767435

f2404:
I see. Then, maybe you could add a patch to your build process that would modify the string for you? I mean, I'm not sure if rpm build allows that but I guess it should.

joequant:
I'd also prefer to run python3, but unfortunately, this requires pyobject3
which hasn't been packaged for Mageia.  It's possible to switch the #! to
python3 and it should still work.  The important thing is to change the
codec code so that it can run with both 2 and 3.

leigh123linux:
@joequant  Are you sure? I see python3-gobject3 listed for Mageia

http://rpm.pbone.net/index.php3/stat/4/idpl/31509157/dir/mageia_cauldron/com/python3-gobject3-3.19.2-1.mga6.x86_64.rpm.html

joequant:
Doh!!!  Let me try that.  What's the ETA for converting all of cinnamon to
python3.  If it's on the schedule, I can close this patch.

joequant:
Closing since I can get the Mageia build working with python3.  Onward and upward!!!

PR #1189, opened 2016-05-28
add standard settings for EelEditableLabel

raveit65:
This set simply a background to this entry, other wise the font is hard to read if you rename a desktop icon, because you see simply the desktop bg.
It's better to set this here as in a theme.
Well, you can play around with the styling details but something like that you need to set.

### DIFF
--- a/src/nemo-style-application.css
+++ b/src/nemo-style-application.css
@@ -15,3 +15,40 @@
     color: #f5f5f5;
     text-shadow: none;
 }
+
+/* EelEditableLabel (icon labels) */
+.nemo-desktop.view .entry,
+.nemo-desktop.view .entry:active,
+.nemo-desktop.view .entry:focus,
+.nemo-desktop.view .entry:backdrop,
+.nemo-window .nemo-window-pane .view .entry,
+.nemo-window .nemo-window-pane .view .entry:active,
+.nemo-window .nemo-window-pane .view .entry:focus,
+.nemo-window .nemo-window-pane .view .entry:backdrop {
+    border-image: none;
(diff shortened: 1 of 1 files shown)


raveit65:
Btw, i did the same for caja.

JosephMcc:
This should be reverted. Themes handle this just fine. We have no issues with our own themes and it doesn't look good at all in some themes.

raveit65:
> This should be reverted. Themes handle this just fine. We have no issues with our own themes and it doesn't look good at all in some themes.

See, what happens with dark themes.
https://github.com/mate-desktop/mate-themes/issues/133#issuecomment-233752577
@JosephMcc @clefebvre

JosephMcc:
@raveit65 Seems fine here: http://imgur.com/a/TINvL
Or am I looking at the wrong thing?

JosephMcc:
@raveit65 One other thing. We do use this line in our themes: https://github.com/linuxmint/mint-y-theme/blob/master/src/gtk-3.0/3.18/sass/_applications.scss#L174
It seems not doing so can cause issues.

raveit65:
@JosephMcc 
Your screenshot shows the right thing, for some reasons the selected bg and fg isn't set up automatic in BlackMATE from mate-themes.
We need to set something there, see my comment in mate report.
I prefer to use the same settings than nemo use already for the desktop.
Let see what Luke want to do, he's the new maintainer for BlackMate.

I don't think that we did set up a transition effect on .entry somewhere in our themes.
So, i am still wondering how **transition: none;** should help.

PR #1210, opened 2016-06-17
[WIP] correct bug causing log error associated with toggling location entry

brownsr:


brownsr:
Unfortunately as exor has pointed out this inadvertently removes the functionality that sets how the user last left the toggle.   And the code setting this is where the error originates from.  More work is needed.

PR #1232, opened 2016-07-22
Fix desktop redraws in GTK 3.21.3 and later

lukefromdc:
GTK 3.21.3 and later broke desktop redrawing except in Nautilus. The commit that made the difference between it working and not working was traced to 
https://github.com/GNOME/nautilus/commit/6d6bf1e7c6049b4bd300ae1d90833a8b5fcf30c4
canvas-container: don't force native window for the desktop canvas 
shortly before Nautilus 3.7.21 came out. To make this work on Nemo also required forcing the scrolledwindow of nemo-desktop-icon-view transparent with a cssprovider, as otherwise it would put a white background on the desktop. Works, running it right now as I post this.

Fixes #1231

PR #1233, opened 2016-07-22
Fix wrong-size desktop in compiz

lukefromdc:
Partially revert 
https://github.com/lukefromdc/nemo/commit/9cb9c48f59baa28a21fa7d5b6f89b4d85d546d4b
which causes the desktop to open on only part of the screen (again) when using compiz.  Using GTK 3.21, had to combine this with https://github.com/linuxmint/nemo/pull/1232 to get a functional desktop on compiz.


clefebvre:
Hi,

Closing the PR after talking to the team:

<exor> the whole reason mtwebster added that was so it works properly with panels on non-primary monitors
<exor> so we could have it iirc
<exor> without htat items go under the panel or something
<exor> it reports the struts or w/e properly for the primary, but for the rest it didn't report them
<exor> i think that was the problem anyways
<exor> https://github.com/linuxmint/nemo/blob/7f53646f9d895d4f0ca78132593eed19510f5722/libnemo-private/nemo-desktop-utils.c#L40
<exor> this one here
<JosephM> that sounds about right
<exor> i think we discussed other window managers maybe not handling it well
<exor> i can't remember what the consensus on that was
<exor> maybe that other DEs would be managing their own desktops? and we had fixed --no-desktop
<exor> i'm not 100% on that part though
<exor> in any case, i don't think we can revert it like that and have it work in cinnamon. and if the desktop window is broken for him then the other one should be too he probably just isn't using it or didn't notice
<exor> the blank one still has functionality, a simple menu, so it needs to be the correct size as well
<exor> long spam:
<exor> I have the following problem: I would like to programmatically change the size of a window and be able to increase its size such that is fills the whole work area on the current monitor. Currently on GTK+ this is only possible for the primary monitor, because the implementation of gdk_screen_get_monitor_workarea does not support multiple monitors. Instead for non primary monitors I can only get the monitor geometry which maybe wrong if there are 
<exor> panels on the secondary monitor. I have discussed this issue in [2], but as I am being told this is a limitation of the underlying EWMH protocol (specifically _NET_WORKAREA), which does not support to query for individual monitors.
<exor> from long proposal here https://lists.freedesktop.org/archives/xdg/2015-October/013576.html
<exor> i mean not that long, but longer than i pasted
<exor> clem an alternate solution to that issue might be "Some WMs just ignore the setting and limit it to its maximized size"
<exor> from this followup msg https://lists.freedesktop.org/archives/xdg/2015-October/013580.html

@mtwebster feel free to review/confirm if you get a chance.

lukefromdc:
I will keep my branch available and continue to build Nemo from it so I have
the ability to use it with Compiz. For my purposes this is needed.

lukefromdc:
If this revert breaks functionality needed in Cinnamon than obviously it could
not be merged. In that case some other fix for Compiz is needed, I was not
sucessful in experiments with having compiz set a fixed size window at the 
screen resolution for this. Any way to do a runtime check to see if the WM is
compiz? 

I'll continue using my branch until a fix is found that does not screw up mulitmonitor
support in Cinnamon, as for what I do with multiple monitors (panel only on primary)
this has never been an issue and I need things to work in other DE's.

PR #1245, opened 2016-08-02
Migrate away from gnome-common deprecated vars and macros

maxyz:
gnome-common is deprecating some vars and macros, listed here:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=829932

This change follows the gnome recommendations from:
https://wiki.gnome.org/Projects/GnomeCommon/Migration

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -38,8 +38,6 @@ DISTCLEANFILES = \
 
 DISTCHECK_CONFIGURE_FLAGS = --disable-update-mimedb --enable-gtk-doc
 
-ACLOCAL_AMFLAGS = -I m4 ${ACLOCAL_FLAGS}
-
 distclean-local:
 	if test "$(srcdir)" = "."; then :; else \
 		rm -f ChangeLog; \
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,13 +1,40 @@
 #!/bin/sh
+# Run this to generate all the initial makefiles, etc.
+test -n "$srcdir" || srcdir=$(dirname "$0")
+test -n "$srcdir" || srcdir=.
 
-srcdir=`dirname $0`
-test -z "$srcdir" && srcdir=.
+olddir=$(pwd)
 
-PKG_NAME="nemo"
+cd $srcdir
 
-which gnome-autogen.sh || {
-    echo "You need to install gnome-common from GNOME Git (or from"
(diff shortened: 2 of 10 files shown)


clefebvre:
dpkg-buildpackage fails with this PR. It looks like warnings are considered errors:

eel-canvas.c:2230:47: error: redundant redeclaration of ‘eel_canvas_accessible_init’ [-Werror=redundant-decls]
 G_DEFINE_TYPE_WITH_CODE (EelCanvasAccessible, eel_canvas_accessible, GTK_TYPE_ACCESSIBLE,
                                               ^

maxyz:
¡Hola Clement!

El 2016-08-17 a las 02:31 -0700, Clement Lefebvre escribió:
> dpkg-buildpackage fails with this PR. It looks like warnings are considered 
> errors:

> eel-canvas.c:2230:47: error: redundant redeclaration of 
> ‘eel_canvas_accessible_init’ [-Werror=redundant-decls]
> G_DEFINE_TYPE_WITH_CODE (EelCanvasAccessible, eel_canvas_accessible, 
> GTK_TYPE_ACCESSIBLE, 
> ^

Indeed, that's the AX_IS_RELEASE([git-directory]) doing, afaik. So, if you 
build it in the git directory it means that you are working in the code and 
could take care of the warnings.

If you don't have a .git directory, then your build is considered from a 
release and then the warnings are still warnings.

The autoconf documentation about these macros can be found:
https://www.gnu.org/software/autoconf-archive/ax_is_release.html
https://www.gnu.org/software/autoconf-archive/ax_compiler_flags.html

You can still use the configure parameter: --enable-compile-warnings=yes to 
avoid this. But I would prefer, recommend to fix these warnings/errors.

Happy hacking,
-- 
"The first 90% of the code accounts for the first 90% of the development time.
The remaining 10% of the code accounts for the other 90% of the development
time."
-- Tom Cargill
Saludos /\/\ /\ >< `/

clefebvre:
Ah cool, it does fix the issue.

Can you add this to your PRs in debian/rules?

We often fix warnings, but it shouldn't stop us from compiling. It's a continuous effort, it's not urgent and shouldn't hold the rest of the work though.

maxyz:
Done

Sorry that it took me so long.

Happy hacking,

clefebvre:
Thanks :)

PR #1257, opened 2016-08-19
Expand grid width to canvas size

ChrisJAllan:
This expands the grid spacing to fit the container, as seen [here](https://gfycat.com/CleverHappygoluckyArawana).

### DIFF
--- a/libnemo-private/nemo-icon-container.c
+++ b/libnemo-private/nemo-icon-container.c
@@ -1339,6 +1339,7 @@ lay_down_icons_horizontal (NemoIconContainer *container,
 	double max_text_width, max_icon_width;
 	int icon_width;
 	int i;
+	int num_columns;
 	GtkAllocation allocation;
 
 	g_assert (NEMO_IS_ICON_CONTAINER (container));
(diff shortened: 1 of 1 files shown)


itzexor:
There is a problem with this commit when you shrink the window/container size smaller than the icon width on max zoom with a folder full of images. The icons become corrupt and a ton of errors are printed to the console. I'm not sure what the problem is at a glance, as it has been some months since I messed with this code. 

If you want, you can check out my own attempt at dynamic grid sizing (among other things) in this branch: https://github.com/itzexor/nemo/commits/icon-view-layout but that has problems of its own that I haven't got around to figuring out yet. It may even have the same issue, I'm not sure if I even tested resizing so small with my branch.

I haven't found a simple way to make it lay them out better without introducing random bugs, as the full layout code is pretty huge and spread across multiple files.

ChrisJAllan:
Ah, yep. It went to zero columns and messed everything up. Fixed that now.

zbandhan:
WoW They fixed it.

itzexor:
I found another weird behavior with this commit. I'm not sure how much of an issue it really is. I only found it because I have a folder of images with one file having a really long filename just for testing icon view.

When the filename is very long, and if the window size is small, the long filename thumb gets pushed off to the right.

before: http://i.imgur.com/uzN68hx.png
after: http://i.imgur.com/tihydLm.png

Issue #569, opened 2014-03-27
unable to set GEOMETRY of nemo


highwindmx:
I am using nemo 2.0.8
nemo has an option --geometry=GEOMETRY , can anyone tell me the right way to write this geometry? I tried:
nemo --geometry=500x500
nemo --geometry="500x500"
nemo --geometry=500x500+50+50
nemo --geometry="500x500+50+50"
but seems nothing(the size and position of the opened windows) changed, 
Why? Or did I misunderstand this option?

Thanks

lukefromdc:
This came in from Nautilus 3.4 from which Nemo was forked. Geometry worked ONLY on the first invocation of the program because of how GApplication works. It was a side effect of an earlier bugfix for opening the wrong folder on "nautilus ./" or other relative paths. That bug was caused by the desktop instance handling the command line when nautilus was invoked a second time and connected to the first instance (one purpose of Gapplication.GtkApplication). They switched to local handling of the command line, but there was no way to get the data from that option from the local to the main instance so setting geometry failed unless Nautilus had not been already started.

Nautilus now uses a more complex way of handling the command line because of this-and at one point had to fix the wrong directory bug a second time. I think they had a need to handle more options, even though they had just removed --geometry.   I am porting Caja to Gapplication/ GtkApplication 
https://github.com/mate-desktop/caja/pull/618
and just ran into this exact same problem. Looks like a hell of a lot of work to port over newer Nautilus code for this, it is very complex. I currently have builds of Caja using Nemo's startup system, so if I get results porting them to Nemo would also be possible.

lukefromdc:
https://github.com/mate-desktop/caja/pull/618/commits/9fb12116c1a09221a88f5c1a7e9010022c96df66
gets --geometry working in my port of Caja to Gapplication using the Nemo/Nautilus 3.4 command line system. For Nemo it can be simplified because the "hint" variable does not have to handle the spatial/browser choice and --browser is ignored, being accepted only for script compatablity.

That "hint" string is one of very few ways to pass option data from a local command line to the main instance, with Nemo it need only handle the geometry string,

I will code this up for Nemo and submit a PR

lukefromdc:
Had to rebase my PR to get rid of unwanted "merge" commits from merging master plus accidental merge to my own master branch while preparing a "dev-all" branch for my own use

PR #1268, opened 2016-08-29
Fix --geometry option when Nemo is already running

lukefromdc:
Fix for https://github.com/linuxmint/nemo/issues/569
 based on
https://github.com/lukefromdc/caja/commit/9fb12116c1a09221a88f5c1a7e9010022c96df66

This fixes a GApplication command line issue inherited from Nautilus 3.4. The issue is that using --geometry in a second invocation of nemo sets self->priv->geometry in the local instance, but the windows are opened by the remote (already running) instance. That reads it's own version of self->priv->geometry, which is not written to by setting the same variable in the local instance. There are very few ways to pass data from a local command line to "open" but Gapplication provides the "hint" string variable. It can pass the geometry string, which can then be written to self->priv->geometry, making the geometry option work again.

### DIFF
--- a/src/nemo-application.c
+++ b/src/nemo-application.c
@@ -533,9 +533,18 @@ nemo_application_open (GApplication *app,
 
 	DEBUG ("Open called on the GApplication instance; %d files", n_files);
 
+	/*Get any geometry string passed by a local (2nd) invocation*/
+	if (strcmp(hint,"") != 0 ){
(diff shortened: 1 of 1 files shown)


lukefromdc:
Very nice cleanup to this, I will see if any of that can be applied to the more
complex case of my Caja branch, where "hint" handles two different options.

PR #1275, opened 2016-09-05
Silence deprecated declarations warnings

leigh123linux:
Do we really need reminding that nemo uses deprecated declarations?


JosephMcc:
I actually don't mind these. While I haven't messed with them in Nemo we often clean some of these up. As long as they are just warnings and not errors.

PR #1284, opened 2016-09-09
Added session manager

sgmanohar:
This session manager simply saves the open tabs when Nemo closes, and reopens them when Nemo opens. 
It doesn't yet preserve multi-pane windows. 
It doesn't have an interface / option to disable the session manager, without recompiling it.


leigh123linux:
This PR has zero chance of being merged and is a total mess!

lestcape:
I'm not consider that this is a mess, just it's a way to add support for unity at the same time that it`s added another feature to nemo. If nemo will be support other desktop or not it's an important decision and not any other feature will be the cause of that... So, like @leigh123linux I recommended split the pr as there are more than one intention here.

clefebvre:
Hi,

I'm closing the PR because it's implementing different features.

I'd like to comment on the main one though, the ability to remember sessions. I think it's pretty cool, but I don't think it's designed correctly. Unlike your web browser, there's a lot more "new window" going on... you get a new window when inserting a removable device, you choose a new window rather than a new tab to tile around the desktop, or to DND between places, so if Nemo was to remember places, I'd want it to remember them all, not just the tabs on my last closed window.

I much prefer the way this is done in Opera for instance.

Have a File->Session->Save opened windows menu item, which opens a dialog and asks for a name. Say I call my session "my session". It remembers all the windows, their position, size, opened tabs, locations.

I can then File->Session->Open "my session" whenever I feel like it.

And with all that said, I'm still not entirely happy with the idea, because I really feel like this should be designed at the WM level so that it can apply to all programs and windows. 

Typically if I save an "activity", it's not just nemo windows I'd want to save, but Firefox, terminals etc.. 

Maybe we need a way for Muffin to be able to extract the path from gnome-terminal, firefox, nemo etc..

PR #1298, opened 2016-09-18
nemo-application.c: fix maximized window state preservation

itzexor:
This commit fixes the maximized window state not being preserved properly since #1268 and the followup commit.  While trying to track this down, I found that we are in some cases passing NULL to g_application_open in the hint parameter which is explicitly forbidden and results in unexpected strings being passed to nemo_application_open. I have added a NULL check there to prevent this.

Aside from that, I also had to add a check for an empty string in open_window otherwise we override the previously restored window state in too many cases.

### DIFF
--- a/src/nemo-application.c
+++ b/src/nemo-application.c
@@ -456,6 +456,7 @@ open_window (NemoApplication *application,
 {
 	NemoWindow *window;
 	gchar *uri;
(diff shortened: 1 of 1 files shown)


itzexor:
I rebased this to add a commit message.

PR #1299, opened 2016-09-18
Make AX_ macros optional

maxyz:
Not all AX_ macros aren't available in previous versions of
autoconf-archive, making them optional allows backports.


daschuer:
I am maintaining a generic version that can be build on e.g. Ubuntu Trusty without cinnamon dependencies. 
https://github.com/daschuer/nemo/tree/generic

My solution was to add the missing scripts to the m4 folder: 
https://github.com/daschuer/nemo/commit/5f789a65d47e55c63a2397c48a6b5f9526b50cd3

PR #1310, opened 2016-10-01
Add option about mouse double click to go to parent folder

claudetete:
I am used to double left click on blank area to go to parent folder on my working computer under MS Windows with QTTabBar (extension of MS Explorer to have tab and useful shortcuts).
So as I started to use Nemo with Cinnamon under ArchLinux, i was expected to found something similar so I propose this patch with an option to not change standard behavior.


clefebvre:
Very nice feature. Thanks.

PR #1321, opened 2016-10-22
file-undo-operations: change trashed files matching condition

leigh123linux:
Based on https://git.gnome.org/browse/nautilus/patch/?id=d1fdc58be0731578572d17fb9c2992b22fab9bbb


clefebvre:
Cool. We should mention the upstream bug report https://bugzilla.gnome.org/show_bug.cgi?id=761819 and upstream developer (here: Razvan Chitu <razvan.ch95@gmail.com>).

We need to formalize this better going forward.

In any case, this is good and should get merged.

PR #1340, opened 2016-11-15
Add missing "Desktop" DEBUG flag

smokku:


clefebvre:
Please explain what this fixes and how to reproduce the issue.

smokku:
If you build Nemo with debugging enabled and you want to see actual debug messages you need to run `nemo` command with `NEMO_DEBUG` environment variable set, enabling debug areas.

For example if you want to see DirectoryView and Window debug messages you run `NEMO_DEBUG=DirectoryView,Window nemo` command which will in turn enable `NEMO_DEBUG_DIRECTORY_VIEW` and `NEMO_DEBUG_WINDOW` debug logs.

Alternatively you may run `NEMO_DEBUG=all nemo` which will enable all (defined) debug keys.

Desktop related code is logging using `NEMO_DEBUG_DESKTOP` but the key to enable the logging messages is missing in current Nemo code, so either `NEMO_DEBUG=Desktop` nor `NEMO_DEBUG=all` works.

Fixing this bug was crucial in enabling me to fix wrong scaling of the Desktop window under Compiz. i.e. code that was logging actual size of the window created was under `NEMO_DEBUG_DESKTOP` flag.

clefebvre:
Thanks :)

Issue #1404, opened 2017-01-31
Nemo hides filename for .desktop files with execute permission


cboxdoerfer:
How to reproduce:

1. Create a file called `malware.desktop`

2. Add the following content to it:

```
[Desktop Entry]
Name=CV.pdf
Exec=sh -c 'touch ~/MALWARE_WAS_HERE'
Terminal=false
Icon=x-office-document
Type=Application
Categories=Office
```

3. Make it executable

Nemo displays the file like that:
![screenshot from 2017-01-31 18-13-29](https://cloud.githubusercontent.com/assets/6108388/22475957/057be074-e7e1-11e6-87ec-deda273a49a7.png)

Once the user opens the file the Exec entry is executed without any confirmation. By hiding the filename and therefore also the filename extension users can easily be tricked to execute arbitrary code when someone ships files like that in an archive which preserves execute permissions.

How to fix it:

Maybe by don't hiding the filename for .desktop files at all.

For reference, this bug also applies to other file managers:
https://github.com/lxde/pcmanfm-qt/issues/449
https://github.com/mate-desktop/caja/issues/727
https://bugzilla.gnome.org/show_bug.cgi?id=777991

clefebvre:
I agree with the analysis done by the pcmanfm team on this. Imo, the right solution is to interpret desktop files as executables and to show a confirmation dialog by default.

I'll close the PR #1405. Please feel free to open a new one if you're interested in implementing the mentioned solution.

ChrisJAllan:
Is PR #1407 enough? It doesn't make it always ask for confirmation, but follow the executable file settings on whether to run, edit, or ask.

ByteHazard:
IMO, only on the desktop view should the *.desktop filename be hidden, and then only with some security checks in place (as suggested in related threads).

Hiding the filename for desktop files opens a can of worms, from security issues (Windows has always had a large variety of exploits, based on a similar Windows feature), to the inability to rename the .desktop files (it requires installing an alternative file manager to change the file extension without resorting to command line trickery, for example).

Moilleadoir:
[...]

PR #1405, opened 2017-02-01
Add option to show filenames for executable .desktop files

ChrisJAllan:
Adds an option to show the filenames for executable .desktop files (default false).

This is motivated both by the security hole hiding filenames causes (see #1404) and by my personal preference.


GitHubIsToxic:
I think this is still a desirable user option, independent of 'treat .desktop files as executable with prompt'. Personally I would prefer that all .desktop files not on the actual desktop should only appear with their proper filenames. Or does this patch clash with those other proposed code changes?

PR #1407, opened 2017-02-02
[WIP] Make executable text file behavior apply to .desktop files

ChrisJAllan:
This makes the executable text behavior (execute, show, or ask) also apply to executable .desktop files.

Because this makes non-executable desktop files always open in the editor, the "Untrusted application launcher" dialogue is never shown. Should this be changed?

### DIFF
--- a/src/nemo-mime-actions.c
+++ b/src/nemo-mime-actions.c
@@ -778,10 +778,6 @@ get_activation_action (NemoFile *file)
 	ActivationAction action;
 	char *activation_uri;
 
-	if (nemo_file_is_nemo_link (file)) {
-		return ACTIVATION_ACTION_LAUNCH_DESKTOP_FILE;
-	}
-	
 	activation_uri = nemo_file_get_activation_uri (file);
 	if (activation_uri == NULL) {
 		activation_uri = nemo_file_get_uri (file);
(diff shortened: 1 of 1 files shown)


mtwebster:
I'm really opposed to taking away the user's control in this matter.  It more or less renders 'trusting' something useless - that's the whole point of it, that you trust it and just want to double-click it and do your thing.

We already allow a person in Cinnamon to right-click some program in their menu and add a trusted desktop file to their desktop, or in nemo, right-clicking a view and creating a new launcher.

I can appreciate the other side though, the potential for something malicious to be unknowingly launched.  I highly prefer gnome's solution to the others, though - I had typed it up, but then I finally looked at gnome's (the other two had depressed me a bit with the attitudes lol) - it is the exact solution I was going to suggest.

The one new thing this brings up, though, is that the metadata database is stored in `~/.local/share/gvfs-metadata` - they're binary files though, I'm not sure how big a problem this is - I haven't looked at how anything is stored in those files.

mtwebster:
I did up a quick patch that adapts what nautilus did to our situation, with the addition of a "View file contents" button to allow the user to check the file out.  I'm not sure having a popup authentication dialog is necessary - perhaps a bit overkill.  But it's an additional option we can add if desired.

https://github.com/mtwebster/nemo/commit/31c4787cf5d80abc988615d855d2c9691e6ae01a

I apologize for waiting this long to review this issue, and for the time you put in on this pull request.  But I'm a lot more satisfied with the outcome if we do it this other way instead - essentially no loss of existing functionality, and it leaves absolute control to the user.  I'm going to try and get this resolved for 3.4 yet.

ChrisJAllan:
My motivation for this is less about security and more about the fact that there is currently no easy way to edit a trusted or executable desktop file, which your patch doesn't address. I see no reason for them to be any different from executable text files in that regard. The behavior is then based on user choice as per their settings.

I do think that the "Untrusted application launcher" dialog should still be shown for untrusted files as to not lose that functionality, so I may add that to this PR.

PR #1431, opened 2017-02-27
nemo-dbus-manager.c: Add MoveURIs method

bluss:
Add MoveURIs, which has the same signature as the existing CopyURIs
method.

Motivation: Better integration with other applications.

Why should a launcher/command shell integrate with nemo? The user gets a more consistent experience, progress, conflict resolution and error handling are all handled by nemo in a familiar interface. The nemo plugin in kupfer is available in its latest version. It can already handle other neat operations that nemo provides:


bluss:
Thanks!

Anyone have thoughts of improvements and further improvements we can work on in this area?

PR #1453, opened 2017-03-24
Connect server dialog

daschuer:
This Adopts the Nautilus Connect to Server Dialog. 

It is really a great improvement in usability. 
It addes also the AFP service (Apple Talk)

connect


leigh123linux:
It looks too dumbed down to be useful, is it still possible to set a custom port for protocols?

> It is really a great improvement in usability.

daschuer:
Yes, It should work. For example 
ftp://ftp.suse.com:80/pub fails as it should but 
ftp://ftp.suse.com:21/pub works 

The improvement is that you can paste the URL you will probably found at the address source, without any knowledge what is which part. Once you have successfully connected you have the new URL in the recent URL list.   

Here is the original discussion from the Nautilus team. 
https://bugzilla.gnome.org/show_bug.cgi?id=682637

But lets discuss how we can improve it based on this? 
Do you have any ideas?

mtwebster:
We can already do this from the location bar.  It accepts server urls and will resolve them and connect - so can nautilus, at least the one in Xenial.. the only thing this brings then is a recent list.  I'd rather keep the existing one with all its controls - it's far better for a less knowledgeable user who may not understand about different protocols, ip addresses, ports, etc...  Sure, it's not as elegant, but it's usable for just about any level of experience with this sort of thing, and I feel strongly that this is an area where, at least for now, usability needs take priority over elegance.

daschuer:
Ah, I understand. 
I think the main issue of the old dialog is that you cannot paste and see the complete URL. Using the address bar instead is not very obvious. 
The single element edit boxes where considered as broken by the Nautilus team, because they did not work well in any case. 

If you think removing them like Nautilus is overdone for Nemo, how about reintroduce them into the dialog proposed here?

The use has always the resulting URL box and we can continually improve the single field edits for odd cases.
We may  make the resulting URL read only an edit it as "edit Url" protocol.

What do you think?

clefebvre:
Hi everyone,

I'll close this PR for now.

@mtwebster @leigh123linux I get the feeling you're not convinced by the design of this dialog.

Please feel free to continue discussing the idea or an eventual different implementation.

I didn't follow this closely but the idea of adding a history to the current dialog without changing its design might be more popular. If you agree on this with one another and you want to lead an implementation, please create a new PR.

PR #1454, opened 2017-03-24
Fix hr strings

muzena:


clefebvre:
no, this should be done in Launchpad.

PR #1455, opened 2017-03-25
Disable selinux for Debian rules

leigh123linux:
Disable selinux for Debian rules or you could add libselinux-dev to the build deps.


daschuer:
This means if one Installs SELinux on Debian, he has to recompile Nemo, to see the SE context. 
Is this intended?

leigh123linux:
Debian doesn't currently compile nemo with selinux support.


https://anonscm.debian.org/git/pkg-cinnamon/cinnamon.git/tree/debian/rules

https://anonscm.debian.org/git/pkg-cinnamon/cinnamon.git/plain/debian/control

Issue #1278, opened 2016-09-05
Italian translation only - Create new Template


Timmy93:
I'm using Linux Mint 17.3 Cinnamon 2.8.8 (but the bug is also in Mint 18) and I think to have found a bug in templates.

I created a model in "~/Modelli" (it's italian) that should be the english "~/Templates".
A simple empty file saved as "text.txt".
Then when I press right click of mouse>Create new document>text.txt it creates me a new file called: "text.txt senza nome" (in english "Unamed text.txt").
The problem is that the only part highlighted is "text" so if I change the name in "newText" I obtain: "newText.txt senza nome".

Is it possible to solve this problem? It's not so heavy to cancel each time " senza nome" from the new file but it's of course a bit annoying.

Here is a picture of the problem.
 ![schermata del 2016-09-05 19 05 57](https://cloud.githubusercontent.com/assets/10899430/18254228/d1f848a0-739b-11e6-86f4-822ef6f0486e.png)

itzexor:
Would "Senza nome text.txt" be incorrect grammar or look bad? If not, then I think the translation needs to be fixed. The template string needs to come after the text for that to work. Currently the string in the Italian translation is "%s senza nome" which is why it has this problem. It will only select up to the file extension which is placed where the %s is.

itzexor:
https://translations.launchpad.net/linuxmint/latest/+pots/nemo/it/960/+translate

Timmy93:
"Senza nome" correspond to "Untitled" in English and is perfectly correct. The problem is that a new document is:
**WHAT IS:** "name.txt senza nome"
**WHAT SHOULD BE:** "name senza nome.txt"

Timmy93:
@itzexor I proposed in launchpad a possible solution: "Nuovo %s", that's the same of "New %s".
In this way there's no need to fix the translation. :)

itzexor:
Alright, cool. I was looking into ways to fix the way we handle that string, but making the translation work is much cleaner.

laltroweb:
Same here (latest mint 64bit italian).
eg > New GIMP Image.xcf senza nome

i glad to entire remove this "senza nome" and maintain just template filename.

any help to solve this s really appreciated :)

Timmy93:
@itzexor I can confirm that also in mint 18.1 the problem is still there. I imagine that my proposal wasn't accepted.

itzexor:
[...]

PR #1463, opened 2017-04-06
[RFC] nemo-file-operations.c: add additional note to translators

itzexor:
The new template document string placeholder '%s' is replaced with the
filename, including the extension. If the placeholder is not at the end of
the string then we will get filenames like "Empty text document.txt Untitled"

See #1278

Edit: either this, or import this patch from nautilus https://github.com/GNOME/nautilus/commit/9d28b3263dbce896bf017438a0059a2da4e32f2c

Let me know.


itzexor:
closing, #1464

Issue #1062, opened 2016-01-01
[Feature Request] Middle click on parent/previous/next folder


hakermania:
Currently, on Linux Mint 17.2. Nemo 2.6.7, when I middle click the upper left buttons ("Go to the previous visited location", "Go to the next visited location", "Go to parent folder") nothing happens (it's the same as trying to move the window by clicking on the title bar).

I think, as Nemo supports tabs, that it would be useful to open the corresponding location on a new tab, as in the Chrome, Chromium and Firefox browsers when you middle click the "Back" and "Forth" history buttons.

PR #1475, opened 2017-05-02
 Middle click on parent/previous/next folder

denis-golubev:
Adjust the event handler of the toolbar buttons "Back", "Forward" and "Up" to handle middle clicks.

The following changes were performed in the file `nemo-navigation-action.c`:
 - `tool_button_press_cb` has been extended to call the new function `new_tab_action`, if the button id is `2` and the event type is "pressed" (otherwise, double clicking with the middle mouse button causes `new_tab_action` to be called three times).
- `new_tab_action`: Similar to `show_menu`, the behavior is dependent on the navigation direction. This function reuses already existing functions from existing event handlers. To open new tabs, the `NEMO_WINDOW_OPEN_FLAG_NEW_TAB` flag is used when calling the aforementioned methods.

Proposed implementation of the feature request #1062.

P.S. This is my first pull request for Nemo. I would be happy about comments on it, e.g. about style (code, commit message) and about amount of documentation (too little / too detailed).

### DIFF
--- a/src/nemo-navigation-action.c
+++ b/src/nemo-navigation-action.c
@@ -185,6 +185,43 @@ schedule_menu_popup_timeout (NemoNavigationAction *self)
                                self);
 }
 
+/* Performs the given action using a new tab, if the selected action
+   supports this. In this case TRUE is returned. */
+static gboolean
+new_tab_action (NemoNavigationAction *self)
+{
+	NemoWindow *window;
+
+	window = self->priv->window;
+
+	switch (self->priv->direction) {
+	case NEMO_NAVIGATION_DIRECTION_FORWARD:
+		nemo_window_back_or_forward (window, FALSE, 1,
+					     NEMO_WINDOW_OPEN_FLAG_NEW_TAB);
+		break;
+	case NEMO_NAVIGATION_DIRECTION_BACK:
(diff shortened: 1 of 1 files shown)


clefebvre:
That's pretty cool. If we follow on the tracks of FF here, could we also support CTRL+Click to do the same thing?

clefebvre:
Not sure we need g_assert_not_reached (); in the default case.

denis-golubev:
Taking a quick look into the documentation, the event structure has modifier states. So it should be possible by just extending the if-condition. I'm going to try it tomorrow.

Initially, I wanted to keep all the unsupported cases out of the switch and just return FALSE in the default case, but the function `show_menu` in the same file did it that way and I tried to imitate the structure. Should I change it?

clefebvre:
I think so, we shouldn't terminate the application or consider an error when it doesn't matter. In fact I wouldn't even enumerate all cases here... I'd enumerate the ones to react on and simply return FALSE in default.

denis-golubev:
Should I create a new pull request for the requested changes of the `new_tab_action` function? I have also added the CTRL-Click behavior.

clefebvre:
hi @denis-golubev, yes please.

PR #1478, opened 2017-05-04
nemo-navigation-action.c: Extend, Fix Middle-Click

denis-golubev:
This pull request contains additions discussed in my last PR #1475.
It extends the Toolbar-Button event handling: CTRL-Clicking also opens new tabs.

While testing, I observed, that opening "Back" and "Forward" in a new tab went to the second element in the history. This PR fixes this also.

Thanks for merging my last pull request.


clefebvre:
Thanks

PR #1482, opened 2017-05-08
Fix some warnings in libnemo-private

XRevan86:


XRevan86:
How did I do things so weirdly… anyway, fixed now.

PR #1501, opened 2017-05-19
Nemo modify desktop icon order

DaveBlack:
Update grid auto arrange icon order. More user friendly due to rubbish bin remaining in the same position, and mirrors the behaviour of other popular OS such as Windows


NikoKrause:
+1

clefebvre:
@mtwebster ?

PR #1513, opened 2017-06-04
List view: Don't scale text size when zooming

leigh123linux:
Based on https://git.gnome.org/browse/nautilus/patch/src/nautilus-list-view.c?id=d055c6f58bdb0cf3a90d6d38ceeef6aec4330c80

When 'list view' zoom is set to 100% the fonts are also scaled which looks horrible.

https://bugzilla.gnome.org/show_bug.cgi?id=308961


clefebvre:
Hi,

I'm not sure I agree with this. When zooming in I'm expecting the text to get larger... and smaller when zooming out.

PR #1563, opened 2017-07-16
nemo-action.h: avoid memory padding

brownsr:
move fields around in structure to avoid memory padding


mtwebster:
rebase please!

brownsr:
rebase playing up on me.  will close, as I suspect there are so few nemoactions flying around that the memory saving is minimal

PR #1598, opened 2017-09-08
Let several file operations work on Google Drive

besser82:
Several fixes for using gDrive with GVFS with Nemo.  Backported from GNOME Nautilus.


Fantu:
thanks, I saw many bug reports about bug/s with rename, hope that this solves them or at least part of them

besser82:
@Fantu All the commits in this PR will solve most of the issues with gDrive (and maybe other GVFS clients).

clefebvre:
I can confirm this fixes being able to undo rename operations.

However I'm seeing a regression on text files... before this PR double-clicking a text file in a Google Drive mount opened it up in Firefox... now, Cinnamon is unable to detect how to open it and asks the user to choose a program for it.

clefebvre:
The regression seems to come from "file: Set activation URI for certain files only".

clefebvre:
I'm unable to test creating new files.. it fails both on master and with this PR.

I would suggest to split the PR so we can merge the functional bits (rename operations, dependencies), and create different PRs for creations and activation URIs so they can be tested and merged in isolation.

mtwebster:
How does one get gdrive working here?  I've setup my account in the new Online Accounts settings.  Is there an accepted/standard way to get drive to show up on gvfs after this?

leigh123linux:
@mtwebster is gvfs-goa installed?

After setting up it should appear in nemo under network.

clefebvre:
rename, undo and glib-dep commits merged.

PR #1609, opened 2017-10-03
Added Czech translation

pavreh:
Translations are taken from Launchpad translations for Nemo.


clefebvre:
Thanks but this file is generated. If these translations are in LP already, they'll make their way into cinnamon-l10n and into that file for the next release.

PR #1610, opened 2017-10-05
nemo-toolbar.c: Add option to hide elevated privileges warning.

### LINKED ISSUE #1200: Add an option to disable Elevated privileges toolbar.
I work with Kali Linux(https://www.kali.org/) which uses root by default. I would thus far really appreciate if the omnipresent bar warning of elevated privileges could be turned off. I would gladly submit a PR if this is deemed doable.

J08nY:
Fixes #1200.

### DIFF
--- a/libnemo-private/nemo-global-preferences.h
+++ b/libnemo-private/nemo-global-preferences.h
@@ -91,6 +91,7 @@ typedef enum
 #define NEMO_PREFERENCES_SHOW_ICON_VIEW_ICON_TOOLBAR   "show-icon-view-icon-toolbar"
 #define NEMO_PREFERENCES_SHOW_LIST_VIEW_ICON_TOOLBAR   "show-list-view-icon-toolbar"
 #define NEMO_PREFERENCES_SHOW_COMPACT_VIEW_ICON_TOOLBAR   "show-compact-view-icon-toolbar"
+#define NEMO_PREFERENCES_SHOW_SUPERUSER_TOOLBAR "show-superuser-toolbar"
 
 /* Which views should be displayed for new windows */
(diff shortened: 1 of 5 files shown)


clefebvre:
This will be renamed show-user-warning. It might be removed in the future if ppl report bugs against it.

Gsettings is user-specific... so this creates issues obviously.

clefebvre:
I'll remove the UI option too.

clefebvre:
https://github.com/linuxmint/nemo/commit/5c31aa6740cf5b7fe13390e11740bad412f69b82

J08nY:
Great, thanks!

StLeoX:
> Nemo version: 5.4.3  

Why removing the UI option?  
So how can I turn off the "elevated privileges" bar excluding rebuild?

mtwebster:
you can try:
```
gsettings set org.nemo.preferences show-root-warning false
sudo gsettings set org.nemo.preferences show-root-warning false
```

This only affects nemo.

advice - 
If you're running root enough that you need to disable the message, I would suggest finding another way to accomplish whatever it is you're doing (if possible) - maybe thru permissions or groups. The entire reason we make it difficult to disable this is because you're one errant keystroke or click away from disaster, and you shouldn't make the mistake of forgetting that.

StLeoX:
I see. That works, thanks.

safir2600:
this warning will persist on xed text editor as @mtwebster said
i couldn't find a setting so i have to delete xed and compile again from source; by deleting manually lines of xed-window.c file which are under: /* If we're running as root, add and infobar to warn about elevated privileges */ between 3401 and 3414

PR #1619, opened 2017-10-14
support vcf files

LohnerZ:
Sorry if I write this wrong place...
It's would be great, if nemo could detect and manage vCard files (or vCard file's symlink).
For example some useful function:
- show contacts on desktop or in a folder with photo
- make a symlink from contacts applications to desktop by drag&drop (app integration)
- show contact details in contacts or mail app (app integration)
- send mail from selected contact (right click menu & mail app integration)
- show received mails from selected contact (mail app integration) 
- send files as attachment with drag&drop event to hovered vCard file 
etc.


leigh123linux:
File your issue to https://github.com/linuxmint/nemo/issues

PR #1625, opened 2017-10-22
send by mail nemo action as core functionality

Curly060:
This action provides a "send by mail" functionality for nemo using xdg-email from xdg-utils.

The file send-by-mail.nemo_action in this PR is identical to https://github.com/linuxmint/mint-common/blob/master/usr/share/nemo/actions/mint-artwork-cinnamon-thunderbird.nemo_action except for:
- Exec directive is now using a local script (send-by-mail.py)
- Dependencies is now xdg-email instead of a hardcoded thunderbird
- Separator is now /#/ (This seems like a "funny" separator, but this form is the only safe one to make splitting the filelist reliable, because / is an invalid character in a UNIX filename and thus /#/ is a 100% reliable delimiter for absolute unix pathes)

The send-by-mail.py script uses Python3 since the demo action suggests that. However, the rest of the Cinnamon suite seems to be using Python2. So maybe it should be Python2 in my script, too?

Part of this PR should also be the removal of this action from the mint-common repository, however, I don't understand the dependencies of the mint-common project.

The reasoning behind this PR is this:
I do believe that "send by mail" should be a core functionality of Nemo. Also, it should not use a hard coded email client, but rather use the preferred mail application, hence my usage of xdg-email. This would make the action in mint-common obsolete.

Any suggestions welcome!

Cheers, Curly060 =;->


clefebvre:
Cool, you're missing the exec permissions on the py script, I'll fix that.

I'll remove the action from mint-common too.

Curly060:
Great, thanks for accepting it! However, are you sure that you deleted everything in mint-common related to this action? I still see a reference to it here:
https://github.com/linuxmint/mint-common/blob/master/generate_additional_files.py

Since I didn't understand where and how this script is being used, I simply mentioned that other stuff needs to be removed, too. So maybe it's worth having a closer look into the mint-common for things relating to this action.

PR #1772, opened 2018-02-05
 Make the libExecPath usable 

leigh123linux:
--libexecdir=/usr/libexec option isn't respected and installs files to  /usr/lib64/nemo/

```
   /usr/lib64/nemo/nemo-convert-metadata
   /usr/lib64/nemo/nemo-extensions-list
```

libexecdir isn't the same as libdir and hard coding nemo isn't acceptable as I want to install in

```
   /usr/libexec/nemo-convert-metadata
   /usr/libexec/nemo-extensions-list
```

### DIFF
--- a/debian/nemo.install
+++ b/debian/nemo.install
@@ -1,6 +1,6 @@
 usr/bin
-usr/lib/*/nemo/nemo-convert-metadata
-usr/lib/*/nemo/nemo-extensions-list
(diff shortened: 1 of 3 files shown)


mtwebster:
Can you adjust debian/nemo.install also (take out the `*/` in the two lines for these files)

PR #1791, opened 2018-02-27
Fix function argument order

Chocobo1:
The order of `gsize max_height` & `gsize max_width` in .c file is different from the header file.


brownsr:
good spot.

PR #1811, opened 2018-03-26
Eject/unmount icon & new folder icon

NikoKrause:
* Add symbolic icon for eject/unmount to the right-click action
* Use folder-new icon in nemos toolbar 
    * The right-click action `Create New Folder` uses the
`folder-new-symbolic` icon.
There's also an option to show `Create New Folder` in nemos toolbar. It
uses `folder-symbolic` icon. Change this to use the same icon as for the
right-click action.


JosephMcc:
Looks good to me.

PR #1848, opened 2018-05-02
Fix some issues with pkgconfig file

Soapux:
Make the extension variable path relative.

### DIFF
--- a/libnemo-extension/meson.build
+++ b/libnemo-extension/meson.build
@@ -70,21 +70,10 @@ if meson.version().version_compare('>=0.41.0')
     name: 'libnemo-extension',
     description: 'A library to create Nemo view extensions',
     version: meson.project_version(),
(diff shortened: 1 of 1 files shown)


clefebvre:
What issues does this fix?

Soapux:
In nemo-3.6.5 the pkg-config file has this as for the variable extensiondir:
extensiondir=${exec_prefix}/lib/nemo/extensions-3.0

In 3.8.1 it looks like this:
extensiondir=/usr/lib/nemo/extensions-3.0

After my change it looks like this:
extensiondir=${libdir}/nemo/extensions-3.0

Also, I simplified the 'libraries' keyword by pointing it to the shared_library nemo_extension_lib so you don't have to explicitly list the link flags.  It produces this result:
Libs: -L${libdir} -lnemo-extension

PR #1964, opened 2018-10-09
Prefer XDS Protocol Type over _NETSCAPE_URL to make DnD work in Firefox.

### LINKED ISSUE #407: Nemo always handles _NETSCAPE_URL  instead of XdndDirectSave
I implemented XdndDirectSave in Firefox, but as long as we support _NETSCAPE_URL, nemo never tries to use XdndDirectSave. XdndDirectSave probably just needs a higher priority.

evilpie:
This fixes #407. I verified that this enables dragging images from Firefox Nightly to this build of nemo.


evilpie:
@mtwebster What is the review process for this repository? Do I need to do anything else?

mtwebster:
Is the same change not required in nemo-tree-view-drag-dest.c for the list view?

Nothing else needs done if that's the case.

clefebvre:
@evilpie ^^

evilpie:
@mtwebster Yes indeed, without that change the list view didn't accept the dragged image. Sorry for the delay.

Issue #1135, opened 2016-04-03
"Move to trash" confirmation dialog option


jez9999:
There should be an option in Preferences -> Behaviour to display a confirmation dialog when delete is pressed and the file would be sent to trash, similar to the dialog when shift-delete is pressed.

I'm aware that the reasoning for not showing this is that sending to trash is a safer operation, but it is still a dangeous operation.  If the file is silently trashed when delete is pressed it is VERY easy for a user to trash a file without noticing.  If they then routinely empty their trash and there are many items there it is very easy to delete the file completely and only notice the file is missing later on, when it's too late.

ElectricRCAircraftGuy:
+1000.

**This is a highly desirable option in my opinion.** If I had 1000 votes to upvote this, I would. I have accidentally hit the delete key while files were selected, and I didn't know where they went or what happened. I also have kids. Weird things can happen, and having the "move to trash confirmation dialogue box" would save a lot of headache.

mohhhh:
**Much needed,** please provide an option for a confirmation dialog when delete is pressed and file(s) moved to trash.

alentsch:
I also missing such a "safety lock" extremely much - please add this option! If a file accidentally gets into the trash-bin because the DEL key is hit without noticing it (that happens to me occasionally) the file risks to be lost forever because who reviews his trash-bin before deleting it? Anything which can go wrong, does go wrong - it is just a question of time. I suspect that many people did not even notice how many files they have already lost due to the lack of this option! Adding this option would not "hurt" anyone, especially if it is off by default. But it would be a great relief for people who have lost files in this way.

rubo77:
The option is already there, but it didn't work: http://unix.stackexchange.com/q/195682/20661

Anything happened on this critical but?

federicocode:
Is extremely dangerous as is it now. I have accidentally delete one file whitout see any notification! Some days later I've empty the trash-bin... and puff, I lose my data. This is really crazy for me. Okay, that option can be not enabled by default.. but the possibility to use an ask button before deleting files is really, really, realy needed.

phavel:
Much needed!

LinuxZoo:
yes - if you have kids / woman - you need this option.

acritum:
# Excuse me, the topic was started 2 years ago and still there is no solution?!! Is it too difficult to add a checkbox in the prefs window? Please add it, it is really very unsafe to delete without a confirmation.

LinuxZoo2:
[...]

PR #2020, opened 2018-12-31
Add Move to Trash confirmation option

devlearner:
First of all, thanks for the Linux Mint 19.1/Nemo 4.0.6 release.
I just switched over from Win10 to Mint Cinnamon, and find that I also missed the "Move to Trash" confirmation (like the others did in #1135).
#### 1. "Move to Trash confirmation" option
So this is my attempt to implement the Move to Trash confirmation dialog: 
move-to-trash-confirm
(which is largely a replicate of the confirm delete dialog)

It's configurable with the first option under Trash in Preferences (default to off): 
> Confirm before moving files to the Trash

trash-options

#### 2. "Use safe defaults" sub-option for deleting files and emptying trash confirmation
While implementing the Move to Trash confirmation dialog above, I came across the following stanza in [GNOME HIG](
https://developer.gnome.org/hig/stable/dialogs.html.en):
> **Default action and escape**
> Assign the return key to activate the primary affirmative button in a dialog (for example Print in a print dialog). This is called the default action, and is indicated by a different visual style. *Do not make a button the default if its action is irreversible, destructive or otherwise inconvenient to the user.* [My emphasis]   If there is no appropriate button to designate as the default button, do not set one.

which makes me wonder whether buttons like *Empty Trash* and *Delete* should be made the default.

Turns out a [question](https://ux.stackexchange.com/questions/33275/which-button-should-get-the-focus-at-delete-dialogs) had been asked squarely on this (tl; dr: the 'answer' suggests in effect the affirmative button is more convenient than 'harmful' in light of the user's intention to initiate the action in the first place).

Still, to me it seems actions like *Empty Trash* and *Delete* are fairly irreversible; (and I didn't seem to find a way to set no default button), so I added a sub-option under "Ask before emptying the Trash or deleting files" to change the default button during confirmation to the negative (like *Cancel*) (so, at the risk of slight inconvenience, requires the user's mousing over or an extra Tab to show intention).

Of course, both the option and the sub-option are optional (and disabled by default).  I've also added a tooltip to the sub-option, to provide some (hopefully) expanatory text.

***

Sorry for the lengthy post.  (This's actually my first time submitting patches.)

Any feedback is welcome!   I'm more than happy to modify or improve the patch based on your feedback.

Thank you very much.

### DIFF
--- a/gresources/nemo-file-management-properties.glade
+++ b/gresources/nemo-file-management-properties.glade
@@ -1240,6 +1240,22 @@
                                         <property name="orientation">vertical</property>
                                         <property name="spacing">6</property>
                                         <child>
+                                          <object class="GtkCheckButton" id="trash_confirm_move_checkbutton">
+                                            <property name="label" translatable="yes">Ask before moving files to the Trash</property>
+                                            <property name="visible">True</property>
+                                            <property name="can_focus">True</property>
+                                            <property name="receives_default">False</property>
+                                            <property name="use_underline">True</property>
+                                            <property name="xalign">0</property>
+                                            <property name="draw_indicator">True</property>
+                                          </object>
+                                          <packing>
+                                            <property name="expand">False</property>
+                                            <property name="fill">False</property>
+                                            <property name="position">0</property>
+                                          </packing>
+                                        </child>
+                                        <child>
                                           <object class="GtkCheckButton" id="trash_confirm_checkbutton">
                                             <property name="label" translatable="yes">Ask before _emptying the Trash or deleting files</property>
                                             <property name="visible">True</property>
@@ -1252,7 +1268,7 @@
                                           <packing>
                                             <property name="expand">False</property>
                                             <property name="fill">False</property>
-                                            <property name="position">0</property>
+                                            <property name="position">1</property>
                                           </packing>
                                         </child>
                                         <child>
@@ -1268,7 +1284,7 @@
                                           <packing>
                                             <property name="expand">False</property>
                                             <property name="fill">False</property>
-                                            <property name="position">1</property>
+                                            <property name="position">2</property>
                                           </packing>
                                         </child>
                                         <child>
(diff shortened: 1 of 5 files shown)


mtwebster:
Hi - apologies for the delay in reviewing this - it seems to work well, I just want to discuss the whole default action thing a bit. 

I think the 'change default option to the contrary' option is a bit excessive to have a preference for, I'd rather we all agree on nice default behavior here for the different situations.  We can always revisit in the future.

Personally, I'd prefer to see the following when confirmations are enabled:
- The dialog for emptying the trash defaults to 'cancel' always.  It's generally not a frequent action, so I think safety can be above convenience.
- The trash and delete actions should default to confirm - if you have confirmations active in the first place, I think the convenience is more important.  Also, to me it makes more sense if both of these dialogs had the same behavior.

Lastly, I'd rather see 'Ask before moving files to the trash' rather than Confirm, to keep it consistent with the delete/empty preference.  (This also reduces inconsistencies with translations)

Thoughts?

Thanks

devlearner:
Hi @mtwebster, thanks for the review (and sorry for the late reply).

Your suggestions make good sense to me.   I agree that'll probably be consistent and sensible defaults for most users.

I've incorporated your feedback as a separate commit (though I think you can do a 'squash and merge' of the pr).

(If you think of other ideas, please feel free to let me know.)

Thanks again for your review!

mtwebster:
Hi, thanks a lot for this.  One thing I noticed as I was looking at the code, please use spaces for indentation in the future.  I know a great deal of nemo's code is tab-based, but we're generally changing that as we perform other changes.

We don't want to go out of our way to convert tabs to spaces either, just around the code you're working on.  (This helps with clarity when looking at commits or bisecting bugs.

Thanks!

devlearner:
Hi @mtwebster, thanks for merging this and the tip.  
Will take note of the formatting tip in the future,

Thanks!

NikoKrause:
I'd like the "Move to trash" button to be red.

Issue #1795, opened 2018-03-04
FEATURE REQUEST: Open multiple tabs via command line


Yoshimaster96:
This feature has been requested as early as 2013. It is now 2018, and it still doesn't exist. I for one would love to have this feature, it would make setting up workspaces via shell scripts (which I do sometimes) much easier.

JosephMcc:
Is there another request for it here?

klemsgithub:
hi, i needed this feature too so i have implemented it :) (https://github.com/linuxmint/nemo/pull/2032)

2635599:
they need to consecrate on having nemo save a group of tabs between sessions first.

klemsgithub:
yes but i have already performed the code modification, all they need to do is to accept the pull resquest i made...

2635599:
that would probably be better as an action you can add. why have scripts and terminal commands to execute when nemo already supports adding actions?

klemsgithub:
I do not understand what are nemo actions. are you talking about context menu?
My need is to launch a nemo windows with a list of specified uri through a script...

icarter09:
Many thanks for submitting your feature request for Nemo. Your feature request has been documented and will be reviewed by the Linux Mint team.

For more information on our workflow and feature requests, read https://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/faq.html.

PR #2032, opened 2019-01-11
Option to open uri in tabs at startup

klemsgithub:
Hello,
This is a patch to allow the opening of multiples uri in tabs at startup thought the command line (usefull to launch a single window with multiples tabs from a script)
As in Caja in which i've already implemented this feature, i've added the -t or --tabs option to open the uri in tabs. Without this option the current behaviour remains (one window is opened per uri)

### DIFF
--- a/docs/nemo.1
+++ b/docs/nemo.1
@@ -42,6 +42,11 @@ Perform a quick set of self-check tests.
 .B \-\-geometry=\fIGEOMETRY\fR
 Create the initial window with the given geometry.
 .TP
+\fB\-t
+.TP
+.B \-\-tabs\fR
+Open URIs in tabs.
+.TP
 .B \-n
 .TP
 .B \-\-no-default-window
--- a/src/nemo-action-config-widget.c
+++ b/src/nemo-action-config-widget.c
@@ -91,8 +91,8 @@ on_check_toggled(GtkWidget *button, ActionProxy *proxy)
 
     g_signal_handler_block (nemo_plugin_preferences, proxy->widget->bl_handler);
     g_settings_set_strv (nemo_plugin_preferences,
-    		             NEMO_PLUGIN_PREFERENCES_DISABLED_ACTIONS,
-						 (const gchar * const *) new_list_ptr);
+                         NEMO_PLUGIN_PREFERENCES_DISABLED_ACTIONS,
+                         (const gchar * const *) new_list_ptr);
     g_signal_handler_unblock (nemo_plugin_preferences, proxy->widget->bl_handler);
 
     g_strfreev (blacklist);
@@ -248,7 +248,7 @@ refresh_widget (NemoActionConfigWidget *widget)
     } else {
         GList *l;
         gchar **blacklist = g_settings_get_strv (nemo_plugin_preferences,
-        		                                 NEMO_PLUGIN_PREFERENCES_DISABLED_ACTIONS);
+                                                 NEMO_PLUGIN_PREFERENCES_DISABLED_ACTIONS);
 
         for (l = widget->actions; l != NULL; l=l->next) {
             ActionProxy *proxy = l->data;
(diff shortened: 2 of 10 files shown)


klemsgithub:
@clefebvre i've make this feature using command line but in the feature request https://github.com/linuxmint/nemo/issues/1795 somebody told me it wasn't a good idea to do so...
It's up to you to decide if you are interested in this modification :)

mtwebster:
Hi, seems to work fine, except for one issue - 

I know in your changes you continue to respect the --geometry option, but if you try to use it in conjunction with --tabs, and supply multiple locations, it fails `do_cmdline_sanity_checks` - thinking geometry will not work with multiple locations.  I don't see why it shouldn't be allowed, though, if we're using the tabs option.

The only other issues I wouldn't mind seeing are style/formatting - you've got a lot of mixed tabs and spaces being used, in new code we use spaces exclusively (though we avoid mass changes of surrounding code from tabs to spaces unless it's a smallish function... best judgement here).

Also, there are some places with no space between function name and open parentheses, and spaces between operators and operands (x > 5 preferred over x>5).  These aren't that big a deal though..

Thanks

klemsgithub:
@mtwebster I've corrected indentation and format. 
So what should i do for the geometry option?

mtwebster:
I meant that you can't provide window geometry (`--geometry`) *and* have use `--tabs` when you provide more than one uri/location.

Add something like this, it should fix it:
```
diff --git a/src/nemo-main-application.c b/src/nemo-main-application.c
index f32b0d0eb..c4d77ccd7 100644
--- a/src/nemo-main-application.c
+++ b/src/nemo-main-application.c
@@ -506,6 +506,7 @@ do_cmdline_sanity_checks (NemoMainApplication *self,
               gboolean perform_self_check,
               gboolean version,
               gboolean kill_shell,
+              gboolean open_in_tabs,
               gchar **remaining)
 {
     gboolean retval = FALSE;
@@ -523,6 +524,7 @@ do_cmdline_sanity_checks (NemoMainApplication *self,
     }
 
     if (self->priv->geometry != NULL &&
+        !open_in_tabs &&
         remaining != NULL && remaining[0] != NULL && remaining[1] != NULL) {
         g_printerr ("%s\n",
                 _("--geometry cannot be used with more than one URI."));
@@ -631,7 +633,7 @@ nemo_main_application_local_command_line (GApplication *application,
     }
 
 	if (!do_cmdline_sanity_checks (self, perform_self_check,
-				       version, kill_shell, remaining)) {
+				       version, kill_shell, open_in_tabs, remaining)) {
 		*exit_status = EXIT_FAILURE;
 		goto out;
 	}
```

Now, the geometry option doesn't actually seem to work currently, but that's another problem.

Thanks

klemsgithub:
done.

clefebvre:
Hi @klemsgithub , 

The indentation changes should not be part of this PR. I know you've been waiting a while for this to get in and it's a bit unfair to request changes that late. Let me know if you're not available and I'll clean this up myself.

klemsgithub:
Hi,
I'm a bit lost and i don't understand what to do...
Should i perform a rollback on the indentation modifications?

clefebvre:
hi @klemsgithub, do you have the indentation changes separate from the actual implementation changes locally? If so, just push without them. If not, checkout the files where the only changes are indentation changes, and cancel the indentation changes in the files still modified.

clefebvre:
Clement, I just noticed another thing.. your email isn't set up within your git config (i.e. we don't see it in the commit history).

klemsgithub:
Ok, i've corrected my git environment with an email.
I don't have any changes not pushed and i've checked all the files i have modified so far and i don't see files with just indentation modifications... Did you saw such a file?

clefebvre:
Sorry, just checking again, you're right... they all have implementation changes.

clefebvre:
@klemsgithub I'll clean it up from here using git show.

klemsgithub:
ok :)

clefebvre:
Ok, it's in https://github.com/linuxmint/nemo/commit/0d2040b56fc0d5ead1c1cbb7a25bf5e5e9fefd86.

I put your name as the committer but I didn't have your email so I left it empty.

If your IDE is fixing the indentation automatically when you save a file you should disable that option. We do fix indentation now and then, but we try to do that in dedicated PRs/commits, and when necessary (in order not to to touch the git file history too much, it makes it easy to git blame that way).

I also noticed a few trailing spaces :) Same thing, don't have them deleted by default for the whole file, but if you have the option in your IDE to have them highlighted, do. That way you can see if you're adding some trailing spaces in your new code and remove them before you commit.

Anyway, thanks for this PR. It's in 4.2.0 now and so it will be part of the next release.

PR #2056, opened 2019-02-05
[WIP] Nemo actions: Add support for exec condition

clefebvre:
The idea is to make it possible for an action to specify a condition which runs a command.

Say we want a "Set as wallpaper" action but only for pictures which dimensions are more than 1000x1000px. We can write a /usr/bin/check-picture-is-big program which calculates the dimensions of the selected picture and returns 0 if it's 1000x1000px or more, or 1 otherwise.

This PR adds the ability for the action to specify:

`Conditions=exec /usr/bin/check-picture-is-big %f`

TODO:

 - [x] Optimize action visibility check to only run exec when everything else matches
 - [x] Support multiple selection
 - [x] Support using %F and all, within the exec condition
 - [x] Review performance and eventual memory leaks
 - [x] Support local path (to be able to place commands/scripts in the same directory as the action file)?

### DIFF
--- a/files/usr/share/nemo/actions/sample.nemo_action
+++ b/files/usr/share/nemo/actions/sample.nemo_action
@@ -3,13 +3,11 @@
 #############################################
 #### DEBUGGING:
 ####
-#### Run Nemo with the environment
-#### variable NEMO_ACTION_VERBOSE
-#### set to get useful log output
-#### for debugging your actions
+#### Run Nemo in debug mode using with
+###  NEMO_DEBUG set to include 'Actions'
 ####
 #### i.e.    $ nemo --quit
-####         $ NEMO_ACTION_VERBOSE=1 nemo
+####         $ NEMO_DEBUG=Actions nemo --debug
 #############################################
 
 # Whether this action is active.  For troubleshooting.
@@ -88,6 +86,8 @@ Extensions=any;
 #     "gsettings <schema> <boolean key>"  is true
 #     "gsettings <schema> <key> <key-type> <[eq|ne|gt|lt]> <value>"
 #     "dbus <name>" exists
+#     "exec <program>" run program and check its exit code (0 is pass, non-0 is fail).
+#                      Enclose in < > if the program resides in the action's folder.
 
 #Conditions=desktop;
 
--- a/libnemo-private/nemo-action.c
+++ b/libnemo-private/nemo-action.c
@@ -1687,20 +1687,122 @@ nemo_action_get_visibility (NemoAction *action,
                             NemoFile *parent,
                             gboolean for_places)
 {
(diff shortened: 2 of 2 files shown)


clefebvre:
With this implemented we can add a few convenient smart actions to nemo to tackle annoying common use cases such as splitting MKVs which are more than 4GB, encoding their DTS audio track into AC3 etc etc.. 

We can make actions which do that already, but as it is now, these actions would show up for any matching mime/extension.. whereas they should really only show up for particular files (file size bigger than 4GB, audio track being DTS..etc).

clefebvre:
I'm testing this with the following action:

```
[Nemo Action]
Active=true
Name=Convert DTS audio to AC3
Comment=Convert the audio track from DTS to AC3
Exec=ffmpeg -i %F -c:v copy -c:a ac3 output.mkv
Icon-Name=xplayer
Selection=s
Extensions=mkv;
Conditions=exec /usr/bin/dts-detect;
EscapeSpaces=true
Terminal=true
```

And here's the condition command (dts-detect):

```
#!/usr/bin/python3

import sys
import os
from pymediainfo import MediaInfo

try:
  path = sys.argv[1]
  if os.path.exists(path):
    info = MediaInfo.parse(path)
    if info is not None and len(info.tracks) > 2:
      audio_codec = info.tracks[2].codec
      print ("CODEC: %s" % audio_codec)
      if audio_codec == "DTS":
        sys.exit(0)
except Exception as e:
  print(e)

sys.exit(1)
```

mtwebster:
I'd like to argue against supporting %F and friends for exec conditions.

Currently we support them in the `Exec` line, but there are numerous options for quoting, escaping, separators.  Would we use those same options for constructing an exec condition?  Or add even more fields so we can define a different construction of the command line for exec conditions?

Keeping it as is lets us just supply file names in the argument vector, ignoring any worries about escaping, etc..  This may cause issues with using existing/bin programs for the exec condition, but I don't think the alternative is worth the effort and complexity.  If your program doesn't work as required, write a simple script to adapt it.

clefebvre:
In the PR I basically hardcoded the fact that the path was always appended at the end. I don't really mind.. I mean if we want to call something that expects it somewhere else, we can always wrap the call with a script so we don't "need" it per se. I just though it'd be more explicit and more flexible if the action was the one specifying where %F was.

smurphos:
@clefebvre - while you are looking at this any chance you could have a peek at #2097

mtwebster:
This is probably ok to merge, I think I've covered all wants.  I apologize for the force push, I had to manually work out some conflicts when I rebased on current master.

clefebvre:
No pb at all, thanks @mtwebster.

PR #2058, opened 2019-02-10
[WIP] Settings inheritance for thumbnailing and view type

Cobinja:
This adds settings to inherit the view type and thumbnailing setting from a parent.
If used, folders inherit those settings from their parent, if said folder does not have an own setting for it.
This also fixes resetting the view.


Cobinja:
Closed in favor of #2094

PR #2094, opened 2019-03-15
Settings inheritance for thumbnailing and view type - Take 2

Cobinja:
This adds settings to inherit the view type and thumbnailing setting from a parent.
If used, folders inherit those settings from their parent, if said folder does not have an own setting for it.
The per-folder-thumbnailing is reimplemented to respect the ignore-metadata setting.
This also fixes resetting the view.


mtwebster:
Can you squash a bit?

Cobinja:
Squashed a bit

PR #2125, opened 2019-04-21
Nemo desktop labels - improved text visibility against light backgrounds

smurphos:
Before
Screenshot from 2019-04-21 08-23-11
Screenshot from 2019-04-21 08-23-25

After
Screenshot from 2019-04-21 08-24-08
Screenshot from 2019-04-21 08-24-27

1366 x 768 screen resolution

### DIFF
--- a/gresources/nemo-style-application.css
+++ b/gresources/nemo-style-application.css
@@ -25,7 +25,7 @@ NemoDesktopWindow GtkPaned {
 
 .nemo-desktop.nemo-canvas-item {
     color: #eeeeee;
-    text-shadow: 1px 1px alpha(black, 0.8);
+    text-shadow: 0px 0px 2px #000, 1px 1px #000, 1px -1px #000, -1px 1px #000, -1px -1px #000, 1px 0px #000, 0px 1px #000, -1px 0px #000, 0px -1px #000;
 }
 
 .nemo-desktop.nemo-canvas-item:hover {


smurphos:
@mtwebster 

I'm wondering as an alternative to this PR if you would consider adding some GUI options to either the Customise Desktop or Desktop settings GUI with maybe 3 or so preset options for Desktop label text shadows - light, medium, and strong or similar and implement that via writing to ~/.config/gtk-3.0/gtk.css in much the same as way you are doing with the scroll bar over-rides?

clefebvre:
Ah yes, that's a bit on the heavy side... I'm not sure it's a better default.

leigh123linux:
I hate the look and would revert it for Fedora if it ever got accepted!

smurphos:
Say it like it is leigh ;-)

User facing options for style or degree of text shadow on the desktop is the way ahead I think. The current default is not satisfactory on light or busy backgrounds.

BrunoNZ:
IMHO it should exist an option to enable at least one solution for this problem, like a grey background, and it doesn't have to be the default setting once that with the default wallpaper the problem doesn't occur.

smurphos:
@BrunoNZ - for a fix in the meantime you can override nemo-desktop's hard coded css via ~/.config/gtk-3.0/gtk.css. You could apply your grey background idea this way as-well.

See https://github.com/smurphos/nemo_actions_and_cinnamon_scripts/blob/master/.config/gtk-3.0/gtk.css for an example. Note the scrollbar width fixes in that file are obsolete for Cinnamon / Nemo 4.2.x as there is now a GUI to do the equivalent.

clefebvre:
Closing due to lack of consensus.

PR #2127, opened 2019-04-25
Open Terminal in parent directory if a file is selected

Tcc100:
A small fix so that the `DirViewActions/OpenInTerminal` shortcut also opens a Terminal if a file is selected. It then uses the parent directory as working directory.

### DIFF
--- a/src/nemo-view.c
+++ b/src/nemo-view.c
@@ -7291,7 +7291,14 @@ action_open_in_terminal_callback(GtkAction *action,
 	view = NEMO_VIEW (callback_data);
 	selection = nemo_view_get_selection (view);
 	if (selection != NULL) {
-        gchar *path = nemo_file_get_path (NEMO_FILE (selection->data));
+        gchar *path;
+        if (nemo_file_is_directory (NEMO_FILE (selection->data))) {
+            path = nemo_file_get_path (NEMO_FILE (selection->data));
+        } else {
+            NemoFile *location = nemo_file_get_parent (NEMO_FILE (selection->data));
+            path = nemo_file_get_path (location);
+            nemo_file_unref (location);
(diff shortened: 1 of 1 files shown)


mtwebster:
This doesn't seem to work for me - I had to change the following as well:

```diff
diff --git a/src/nemo-view.c b/src/nemo-view.c
index 221a7ec35..53fda839e 100644
--- a/src/nemo-view.c
+++ b/src/nemo-view.c
@@ -9687,7 +9687,8 @@ real_update_menus (NemoView *view)
 
     action = gtk_action_group_get_action (view->details->dir_action_group,
                                          NEMO_ACTION_OPEN_IN_TERMINAL);
-    gtk_action_set_visible (action, no_selection_or_one_dir);
+
+    gtk_action_set_visible (action, selection_count <= 1);
 
 	action = gtk_action_group_get_action (view->details->dir_action_group,
 					      NEMO_ACTION_NEW_FOLDER);
```

Otherwise the Open In Terminal action still only shows when a single directory is selected or it's a background (no selection) click.  If you want to add something like that then I can merge it.

Thanks!

Tcc100:
I only use this with a keyboard shortcut like this:
`(gtk_accel_path "<Actions>/DirViewActions/OpenInTerminal" "<Alt>t")`

I didn't consider it that useful in the files' context menu, because when I'm using the mouse, I also can click a folder. So I didn't want to further fill this menu with seldom used entries.

Also the label _Open In Terminal_ might be misleading for some files, especially Shell Scripts.

If you think this entry is beneficial for many users, I'll add it.

mtwebster:
Ah good points, and yeah there really are enough items in the context menus :)

Never mind, then.

Issue #2128, opened 2019-04-28
sort order


damian101:
I always use underscores and sometimes exclamation marks to make specific folders appear at the top of the list when sorting by name. That worked in the Explorer on Microsoft Windows and works with Dolphin and the Deepin File Manager on Linux. But in most other file managers on Linux, including Nemo, this doesn't work, because they sort files in a different way. Most Nautilus-based file managers just take the "`ls`-sort-order" and apply natural sorting to it.
I would like to see a sort order similar to what is used in Dolphin and Deepin File Manager, as that is what people are expecting and Windows users are used to. 
The "`ls`-sort-order" also doesn't make much sense normally and might change with distribution and system language.

Meru:
https://github.com/linuxmint/nemo/issues/2085

xenopeek:
Nemo converts names to collation keys before sorting, using functions from https://gitlab.gnome.org/GNOME/glib/-/blob/master/glib/gunicollate.c. Among other reasons the purpose of this is natural number sorting e.g., so that "42towels" is sorted before "1024towels". Downside of this as you've noted is that you can't easily make a directory sort at the the top of the list by prefixing it with some character. As noted you can achieve that instead by pinning it (from the context menu or the Edit menu select Pin, or press Shift+Ctrl+D). 

There is a more detailed discussion in #2135 about Nemo's sorting and possible additional options for that.

PR #2145, opened 2019-05-26
[WIP] Allow pinning files to the top of a file list

mtwebster:
This is an attempt to implement a way to keep important files/folders at the top of a view.

For now, there's a Pin/Unpin item in the expanded context menu.  Multiple selection is allowed, but the first item in a file list determines the action (pin or unpin) regardless of the state of the remaining items.

Currently using a star emblem in icon view, boldface rows in list view.  This may change.

Pinned files will *always* appear at the top of the file list, regardless of sort direction.  They will, however, follow sorting rules amongst themselves.

ref #2128
image (1)
image


mikeroyal:
This looks really good @mtwebster. I'm not sure about the star icon an idea I have would be to use a green check-mark icon. Provided an example screenshot below.
screenshot

mtwebster:
I made a 'pin' emblem, but more similar to other standard emblems currently used (like the lock symbol for read-only files, etc..)

This is definitely something that can be changed/tweaked.  My inkscape skills are shaky at best. :)
image

Issue #2157, opened 2019-06-09
Guake does not open in current folder if set as default terminal


BeatLink:
```
 * Nemo version (nemo --version): nemo 4.0.6
 * Is issue with desktop or windowed nemo?: Both
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...): Mint Cinnamon 19.1
 * Graphics hardware *and* driver used: Intel Corporation 2nd Generation Core Processor Family Integrated Graphics Controller
 * 32 or 64 bit: 64bit
 ```

**Issue**
If Guake is installed and set as the default terminal, opening it using the context menu or toolbar item will result in it opening in the home folder rather than the current folder

**Steps to reproduce**
Download and install guake
Set it as the preffered application using cinnamon settings
Open nemo
navigate to a folder that is not the home folder
right click and click open terminal here

**Expected behaviour**
Guake should pop up, with the current folder as the working directory

**Other information**
This is due to the fact that guake requires arguments while opening. Specifically, the following command `guake --show -n %P` will create a new tab with the current folder as the working directory, and then pop up the interface itself. The issue is, the preffered applications ui only sets `guake` as the execution command, without the arguments. A temporary fix for the time being could be to bring back the "exec-arg" key in Gsettings under org.cinnamon.desktop.default-applications.terminal and have nemo use it.

BeatLink:
There was once a commit that actually implemented this feature https://github.com/linuxmint/nemo/commit/7ca7990c0c33fa4a4f19ba0f534f00f221acb480 however it was reverted presumably due to lack of need https://github.com/linuxmint/nemo/commit/d18a44c67c7385cae8d2ec40c8d625797e6a411e

however, reinstating this commit would solve the issue.

BeatLink:
Until this issue is fixed, I've taken the liberty of creating my own script which reads the current directory and launches guake with the necessary command. Bonus point is the script is self configurable by running it with the --setup argument

https://gitlab.com/snippets/1866239

PR #2158, opened 2019-06-09
Fix issue 2157 (Support the exec-arg gsettings for the open in terminal command)

BeatLink:
This is a recreation of https://github.com/linuxmint/nemo/commit/7ca7990c0c33fa4a4f19ba0f534f00f221acb480

### DIFF
--- a/libnemo-private/nemo-global-preferences.h
+++ b/libnemo-private/nemo-global-preferences.h
@@ -237,6 +237,7 @@ typedef enum
 
 /* Terminal */
 #define GNOME_DESKTOP_TERMINAL_EXEC        "exec"
+#define GNOME_DESKTOP_TERMINAL_EXEC_ARG    "exec-arg"
 
 /* Tooltips */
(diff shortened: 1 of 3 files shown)


JosephMcc:
Thanks for the contribution but it would be nice to see the commit message actually say what this does. When someone is coming back to this in the future it is much more clear that way. This could be as simple as "support the exec-arg for the open in terminal command"

BeatLink:
Fixed! (Though I had to force push the amended commit)

clefebvre:
Why bring this commit back?

It was reverted and the justification for this is described at https://github.com/linuxmint/nemo/commit/d18a44c67c7385cae8d2ec40c8d625797e6a411e.

BeatLink:
I already mentioned that guake cannot work without arguments

clefebvre:
https://github.com/linuxmint/nemo/issues/2157

BeatLink:
Yes that's the one

Issue #2214, opened 2019-10-01
Bug: Pango-1.44 hyphenates  long file names.


prashantraina:
```
 * Nemo version 4.2.3
 * Desktop and windowed nemo.
 * Distribution - Manjaro (Arch)
 * NVIDIA RTX 2080Ti, NVIDIA driver version 430.40
 * 64 bit
 ```

**Issue**

Since a recent update, long filenames are being hyphenated as shown in the screenshot.
This is very confusing, because the filenames may contain actual hyphens.
The non-existent hyphens are still visible even when renaming files! It causes confusion when modifying the file name.

![image](https://user-images.githubusercontent.com/9632639/66007349-06a60080-e480-11e9-90ad-577bf4138fa5.png)

**Steps to reproduce**
Create a file with a long name (> 8 characters on my machine, but this might depend on screen resolution).
Also, when renaming the file using the context menu option, the hyphens are still visible.

**Expected behaviour**

Please add an option in the Preferences or somewhere to disable hyphenation of long filenames and use the old behavior for wrapping long filenames.

**Other information**

Salamandar:
My duplicate issue was closed so I'm here to support this issue :)

icarter09:
Many thanks for submitting your feature request for Nemo. Your feature request has been documented and will be reviewed by the Linux Mint team.

For more information on our workflow and feature requests, read https://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/faq.html.

Salamandar:
That's not a feature request, but a "revert feature" request.

PR #2257, opened 2019-12-08
nemo-icon-canvas-item.c: Don't hyphenate long filenames

leigh123linux:
Fixes for https://github.com/linuxmint/nemo/issues/2214

Pango 1.44 got the ability to automatically hyphenate on line breaks,
which is on by default, but can be set off by a new attribute.

As a result, we now hyphenate filenames, which is confusing, because
a filename may already include hyphens.

To restore the previous behavior, let's not insert hyphens when
breaking filenames in multiple lines.

Based on https://github.com/mate-desktop/caja/commit/b50e36b639aca25c8bf3d21dc1547fc6f7a6f442

### DIFF
--- a/eel/eel-editable-label.c
+++ b/eel/eel-editable-label.c
@@ -36,6 +36,15 @@
 #include <gtk/gtk.h>
 #include <gdk/gdkkeysyms.h>
 
+#ifndef PANGO_CHECK_VERSION
+#define PANGO_CHECK_VERSION(major, minor, micro)                          \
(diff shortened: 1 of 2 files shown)


leigh123linux:
@eli-schwartz Arch requires this.

Salamandar:
\o/

eli-schwartz:
Thanks, Arch is indeed affected. Is this the final version of the patchset and ready to merge?

Issue #2177, opened 2019-07-23
Sort hidden files the same as normal files


stevenxxiu:
```
 * Nemo version (nemo --version) 4.2.1-1 on Arch Linux
 * Is issue with desktop or windowed nemo? Desktop
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) Arch Linux
 ```

**Issue**
Hidden files (those beginning with `.`) are always sorted later, after regular files when they are shown.

They should be sorted the same as regular files, and not sorted after regular files. For example, on both Mac and Windows, they are sorted the same way.

This could be an option to have them sorted later if really needed, but I find this rather strange.

Sorting them in the same way probably reduces the amount of code too.

I also made a comment [here](https://github.com/linuxmint/nemo/issues/578#issuecomment-500096351).

mtwebster:
From left to right:
Nemo, Nautilus, Caja, Thunar

![image](https://user-images.githubusercontent.com/262776/72289972-3ab1c580-361a-11ea-8a57-e33221922e6f.png)

All of them sort hidden files separately from normal files.  The only difference is thunar sorts hidden files *before* while the other three sort them *after* normal files.

None of them offer an option to change this, and I don't really think it would be an improvement (nor do others I've discussed this with).  I find it easier to identify hidden folders/files the way it is now, regardless of how windows or mac does it.

Issue #578, opened 2014-04-03
[Feature Request] Pure ASCII filename sorting


lore68cnc:
Nemo 'by name' files sorting uses 'logical ordering' which fails in some cases (always with hexadecimal character in filenames).
I request to introduce an option in Nemo Preferences -> Behavior to use a "pure ASCII" order, like "ls -la" output.
In Windows this is very simple, see http://support.microsoft.com/kb/319827.
Why Linux is more more complicated than Windows?

Related to "Sorting by filenames - Incorrect order" #205

NETHeader:
+1
I copy bbshopadmin's statement, would love to see that feature in nemo.

hachesilva:
+1

ByteHazard:
+1
Please fix this. This really is shortcoming so important, it should be classified a bug. There are no excuses, Windows 3.11 did this 1994!

bernd-wechner:
+1 as well. Danged frustrating smarty sorting. Give us some sort options here, with plain old ASCII as one. Give me same order as `ls -l` for example.

Vahan86:
@JosephMcc, can we mark this issue as a feature request?

baikalov:
+1 
Is there an ETA for this? This is really frustrating to use it without ASCII sorting.
Please fix this!

bernd-wechner:
Please, please, please fix this!

binarez:
Is this caused by g_uri_escape_string ? It seems like the filenames are passed thru this func before sorting. Not sure if changing this would be a security issue. Thoughts?

ManIVIctorious:
The file sorting is based on your local settings, I don't think this is in the scope of the project...

bernd-wechner:
> The file sorting is based on your local settings, I don't think this is in the scope of the project...

I disagree. To begin with ls doesn't do any such thing, and I'd expect the default behaviour of a GUI file browser to be the same as ls. Secondly it is perfectly valid to desire a different sorting mechanism for different apps on a desktop system and so implementing a sort setting in Nemo one selection on which might be "As defined by locale" is sensible.

But that aside, you might help us by explaining just which locale settings influence Nemo's sorting behaviour and how to set them.

ManIVIctorious:
Sorry, forgot about this...
The environment variable is LC_COLLATE (in my case =C) and nemo behaves exactly as ls does on my system

bernd-wechner:
Nice. It does come with caveats mind you:

[...]

PR #2260, opened 2019-12-09
Sort hidden files the same as normal files

stevenxxiu:
Sort hidden files the same as normal files, so we don't have special sorting for `.` and `#`.

This lets Nemo be the same as other Linux file managers (e.g. Thunar), Windows explorer, Mac file manager, and `ls`.

See issues:
- https://github.com/linuxmint/nemo/issues/2177
- https://github.com/linuxmint/nemo/issues/578#issuecomment-500096351

### DIFF
--- a/libnemo-private/nemo-file.c
+++ b/libnemo-private/nemo-file.c
@@ -93,10 +93,6 @@
 #define DEBUG_REF_PRINTF printf
 #endif
 
-/* Files that start with these characters sort after files that don't. */
-#define SORT_LAST_CHAR1 '.'
-#define SORT_LAST_CHAR2 '#'
-
 /* Name of Nemo trash directories */
 #define TRASH_DIRECTORY_NAME ".Trash"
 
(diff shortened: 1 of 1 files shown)


stevenxxiu:
Someone want to have a look at this? Really need this feature, and it's quite a small patch.

stevenxxiu:
@mtwebster Sorry for messaging directly, but could you have a look at this? Currently I'm compiling my own nemo with my patch included.

I can make any modifications/additions you require.

Really want this to be included as the other sorting issues can partially be addressed by a special locale as discussed in https://github.com/linuxmint/nemo/issues/578, but this one's hard-coded.

mtwebster:
https://github.com/linuxmint/nemo/issues/2177#issuecomment-573860518

bernd-wechner:
I'm mildly confused. When I do `ls -a1` I see all the hidden files sorted at top (`.` sorting higher than anything else). I'm quite used to that and like it that way. Of course I do think Nemo should have options for these sort preferences so that we can all be happy!

PR #2292, opened 2020-01-05
test using upstream's gnome-desktop thumbnailer

mtwebster:


leigh123linux:
It compiles ok here with gnome-desktop3-devel-3.34.2-1.fc31.x86_64

PR #2310, opened 2020-01-27
Added create folder with selection

pedro-augusto-santana:
Added the "create folder with selection" to the default set of Nemo actions, from my [repo](https://github.com/pedro-augusto-santana/nemo-create-folder-selection), only added the files in files/usr/share/nemo/actions, any errors or suggestions please post here so I can try and fix it

### DIFF
--- a/files/usr/share/nemo/actions/askpass.sh
+++ b/files/usr/share/nemo/actions/askpass.sh
@@ -1,0 +1,3 @@
+#! /bin/bash
+
+zenity --password --width 450 --title="Authenticate" --text="Enter your root password"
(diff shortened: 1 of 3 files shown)


dhonx:
Maybe the first feedback is an undo handling feature. When we move something & then hit <kbd>ctrl</kbd> + <kbd>z</kbd>, them the items will back to their place before to this action.

mtwebster:
Hi, thanks for this,  but I think I'd rather bake this directly into nemo - it's a nice feature to have and I'd prefer the 'new folder naming' workflow stay consistent with the rest of nemo (along with having undo support).  I'll add to our roadmap.

pedro-augusto-santana:
Yeah, I made this for myself more as a utility, I posted it on github just to make available to anyone who wants it. Looking forward to seeing it implemented in nemo. Have a nice day!

Issue #2065, opened 2019-02-15
tracker 2.0 API is not supported


leio:
```
 * Nemo version 4.0.6
 * Distribution - Gentoo
 * 64 bit
 ```

**Issue**

Tracker API 2.0 is unsupported

**Steps to reproduce**

Try to compile nemo on a system with tracker-2 or newer with `-Dtracker=true`. It'll fail configure.

**Expected behaviour**

tracker-2 works

**Other information**

It is suprising that after all this time nemo still doesn't support tracker-2.0 API. Should tracker support be hard-disabled and not provided as an option, given how unloved it seems?

leio:
https://gitweb.gentoo.org/repo/gentoo.git/tree/gnome-extra/nemo/files/tracker-2.patch fixes the build for me; runtime completely untested by me

leigh123linux:
If you want it please submit a pull request.

leio:
Why close unresolved issues?

leigh123linux:
Submit a pull request if you want it!!

https://github.com/linuxmint/nemo/pulls

leigh123linux:
> Why close unresolved issues?

Are you to lazy to submit it?

leio:
With such an attitude, most definitely lets say I'm too lazy.

I maintain tracker downstream, and merely had to make some software like nemo work with a newer version, so that I wouldn't break it. I have no interest in nemo itself whatsoever. I even provided the link to a patch I'm using 10 months ago. A PR to me would mean I actually tested it at runtime. Though our users seem to have been, in terms of not filing any issues downstream.

So, no, I don't want it. You should though. Otherwise distributions keep patching your software or decide it's too badly maintained to keep it packaged.

leigh123linux:
Linuxmint, Archlinux, Debian and Fedora don't enable it.
I have zero interest in doing runtime testing as tracker hogs my cpu and spams the logging.

@mtwebster @eli-schwartz @maxyz Are you interested in this untested patch?

clefebvre:
Well it is a build option we provide, so technically yes, personally I'd consider this a bug.

The patch is also tiny and whatever runtime testing is involved, it can't be worse than a build failure.

@leigh123linux you're probably reacting to "suprising that after all this time" and "given how unloved it seems". "lazy" and "if you want it" surely won't help either though.

I've no interest in this personally, but I agree it is a bug and we should either support tracker 2, or remove that option entirely.

leio:
[...]

Issue #1551, opened 2017-07-10
nemo 3.4.6 segmentation fault on search:  'NEMO_IS_SEARCH_ENGINE (engine)' failed


mmirg:
```
 * Nemo version 3.4.6
 * Is issue with desktop or windowed nemo? windowed
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) Gentoo/Funtoo
 * Intel with xf86-video-intel 2.99.917_p20170216
 * 64 bit
 ```

**Issue**
Nemo often segfaults on search as referred to in the comments in bug #1546 but this crash occurs even when patched with 2bee2d5.

**Steps to reproduce**
I can't determine an exact set of steps to reproduce, but one way in which I can relatively reliably crash nemo is by creating a new folder with an arbitrary file in it, browsing to the folder in nemo and then searching for the filename of the new file.

**Expected behaviour**
Nemo will display the file in the search results and not crash.

**Other information**
Out put from NEMO_ACTION_VERBOSE=1 nemo

(nemo:24284): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'NemoSearchEngineTracker'

(nemo:24284): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'NemoSearchEngine'

** (nemo:24284): CRITICAL **: nemo_search_engine_finished: assertion 'NEMO_IS_SEARCH_ENGINE (engine)' failed
Segmentation fault

mtwebster:
Hmm can't reproduce here on Mint in any form.

Ah I also just noticed you're possibly using tracker, that could explain the different behavior.  We don't compile tracker support normally.  I'll see if i can get a chance to test it out.

mmirg:
Ah yes, tracker was enabled not so much by choice as that it was enabled by default.  Disabling tracker support when compiling nemo seems to solve this issue.

leigh123linux:
This is fixed in the latest nemo version.

PR #2334, opened 2020-03-20
Add tracker support

J08nY:
This PR reverts 5db7a44e8086d1b77f6384b272353c932404a05a and applies a patch to add support for `tracker-2` taken from #2065 

There is really no reason to drop support for this when there is a patch that applies cleanly and keeps  it usable with current versions of tracker.

### DIFF
--- a/libnemo-private/meson.build
+++ b/libnemo-private/meson.build
@@ -91,6 +91,11 @@ if libexif_enabled
   nemo_private_deps += libexif
 endif
 
+if tracker_enabled
+  nemo_private_sources += 'nemo-search-engine-tracker.c'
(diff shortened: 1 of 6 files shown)


leigh123linux:
-1 to this request, it doesn't address any of the previous tracker issues .

J08nY:
> -1 to this request, it doesn't address any of the previous tracker issues .

I found #1551 that seems related, are there any other issues that you know of? Will look at this one.

J08nY:
I was not able to reproduce #1551, I built this branch and tested the provided example with tracker both on and off and it did not trigger the segfault.

PR #2341, opened 2020-03-24
[wip] Performance improvements

mtwebster:
Use a static spacing for the icon layout, this eliminates constant re-layouts as thumbnails are loaded.

Delay thumbnail loading initially, until the view is finished being populated.  This is by far the costliest async operation, and when loading a pre-existing thumbnail ends up having a large impact on view loading time.  
The file may already have a thumbnail (if it was spawned by another folder or tab), in this case we delay using that thumbnail until the end.

Once the view is fully populated, the currently visible items have their thumbnails loaded.  As the view is scrolled, the thumbnails will be loaded on demand.


mtwebster:
changes in master now.

PR #2361, opened 2020-04-24
Add %e action token (extension stripped filename)

diath:
The new modifier will return the file name with the extension stripped.
This allows us to create actions such as "Extract \<archive.ext\> to \<archive\>..." to avoid a situation where common "Extract here" actions would otherwise dump contents of the archive to the current directory if it does not contain a root directory inside while being quicker to use than "Extract to" actions, for example:
```[Nemo Action]
Active=true
Name=Extract to %e...
Comment=Extract "%f" to "%e"
Exec=file-roller --extract-to=%P/%e %F
Icon-Name=gnome-mime-application-x-compress
Selection=notnone
Extensions=zip;7z;ar;cbz;cpio;exe;iso;jar;tar;tar.Z;tar.bz2;tar.gz;tar.lz;tar.lzma;tar.xz;
Quote=double
```

### DIFF
--- a/files/usr/share/nemo/actions/sample.nemo_action
+++ b/files/usr/share/nemo/actions/sample.nemo_action
@@ -22,7 +22,7 @@ Active=false
 # %f or %N (deprecated) - insert display name of first selected file
 # %p - insert display name of parent directory
 # %D - insert device path of file (i.e. /dev/sdb1)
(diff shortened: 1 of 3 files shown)


mtwebster:
Nice, thanks.

PR #2363, opened 2020-04-26
Add places sidebar preferences

diath:
This adds customization preferences to toggle the following items from the places sidebar:
* Recent
* Trash
* Bookmarks group
* Network group

There's no GUI way to toggle them as I'm not familiar with Glade however they can be toggled using gsetting for the time being:
* `gsettings set org.nemo.preferences places-show-recent (true|false)`
* `gsettings set org.nemo.preferences places-show-trash (true|false)`
* `gsettings set org.nemo.preferences places-show-bookmarks (true|false)`
* `gsettings set org.nemo.preferences places-show-network (true|false)`

### DIFF
--- a/libnemo-private/nemo-global-preferences.h
+++ b/libnemo-private/nemo-global-preferences.h
@@ -64,6 +64,10 @@ typedef enum
 
 /* Sidebar panels  */
 #define NEMO_PREFERENCES_TREE_SHOW_ONLY_DIRECTORIES         "show-only-directories"
+#define NEMO_PREFERENCES_PLACES_SHOW_RECENT                 "places-show-recent"
+#define NEMO_PREFERENCES_PLACES_SHOW_TRASH                  "places-show-trash"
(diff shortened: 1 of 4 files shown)


mtwebster:
Hi, I don't think we need more preferences here, these are pretty specific, and there's already functionality to clean up the sidebar of things you don't use often:

- You can already collapse/expand sections and their 'last' states are stored in preferences.
- Bookmarks can be placed in the My Computer section as well as Bookmarks.  If you turn off Bookmarks, you'd possibly still have some visible under My Computer, and newly created bookmarks are placed under Bookmarks, not My Computer.
- The recent files turns off if recent files are disabled for the system. (Privacy settings for cinnamon).

diath:
I see, while they are very specific, the rationale behind them was to hide some clutter from the interface when things are not in use since most other interface elements can already be toggled (either through the views menu or preferences window).
I was mostly hoping that I could get this merged upstream so that I don't have to manually patch every new release but if it's not something you want that's completely fine. Keep up the good work :+1:.

PR #2364, opened 2020-04-26
[WIP] Show tab area permanently if desired

wschenck:
In this PR a new setting is introduced ("Always show tab area") in the preferences dialogue box under "Behavior". If this checkbox is marked, all panes always show the tab area even if there is only one tab.

Two downfalls:
(1) On my current system, with the library packages provided by the distribution (Ubuntu 18.04), I only can build nemo 3.8, therefore the PR is not tested with master and does not merge cleanly with master.
(2) Right now, changes in the new setting only become effective for new panes, not for the already existing ones (however, add/remove tab from pane also causes an update). This could be improved by a callback function which updates all panes after a change to this setting was made (on closing the preferences dialogue box). Unfortunately, I don't have the time to dig that deeply in the code basis of nemo right now (I am also not a GTK expert), therefore I have decided to just create this PR as a starting point for further discussion or code development in this area.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,12 @@
+nemo (3.8.6) tara; urgency=medium
+
+  [ Michael Webster ]
+  * nemo-icon-canvas-item.c: Insert some space between the icon and label bounds in compact view.
+  * nemo-query-editor.c: don't intercept keys when hidden.
+  * nemo-places-sidebar.c: Fix a few issues with mounting/ejecting - some incorrect _finish functions and some assumptions of error always being set.
+
+ -- Clement Lefebvre <root@linuxmint.com>  Tue, 11 Sep 2018 14:52:55 +0100
+
 nemo (3.8.5) tara; urgency=medium
 
   [ Michael Webster ]
--- a/gresources/nemo-file-management-properties.glade
+++ b/gresources/nemo-file-management-properties.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.20.2 -->
+<!-- Generated with glade 3.22.1 -->
 <interface>
   <requires lib="gtk+" version="3.10"/>
   <object class="GtkListStore" id="model1">
@@ -260,6 +260,9 @@
     <property name="default_width">800</property>
     <property name="default_height">600</property>
     <property name="type_hint">dialog</property>
+    <child>
+      <placeholder/>
+    </child>
     <child internal-child="vbox">
       <object class="GtkBox" id="dialog-vbox1">
         <property name="visible">True</property>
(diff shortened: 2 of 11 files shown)


mtwebster:
Hi, I've been thinking about doing this for some time, it just hasn't been a priority.

This seems to work as intended.  I had hoped to be able to have a new tab button also to go with this - so, essentially it appear like a web browser with the + new tab button always visible.  I'm not really certain of its usefulness in its current state.

I'm going to keep this open but tag for next dev cycle (we're at the end of this one, and this isn't a quick fix to implement fully).

Thanks

wschenck:
Thanks for the feedback. I have continued to work a little bit on this but found another problem. Right now - when one single tab is made visible -, it is possible to drag this single tab out of the notebook. This can have a lot of funny effects when more drags follow back and forth. 

To counteract this, I have tried to implement proper closing of a single pane or of the whole window when no tab is left there but so far I was not successful. All my solutions (based on so far incomplete knowledge of the nemo code design and of all the possible interactions between components) led to a seg.fault eventually...

wschenck:
I have added some further commits as suggestions.

(1) I like the idea of double-clicking on tabs to directly duplicate this tab. This was implemented and would in my view reduce the need for a "new tab" button directly in the tab bar. But of course, this is a matter of taste and maybe of the overall GUI philosophy in Cinnamon which I don't know.

(2) As described earlier, I did not succeed in proper closing of windows with empty notebooks. As a intermediate solution, I implemented that a single tab in a notebook is non-detachable. As soon as there are at least two tabs, they are detachable again. 

(3) The implementation of feature 2 is really straightforward, however, as soon as a new window is created by dropping a tab to the desktop, this new tab is somewhere in a very secret place set to "detachable" again... I really did a lot of debugging, but in the end this stayed a mysterious phenomenon. Therefore I created a workaround (commit 706bb0a) which transfers the location in the dropped tab (slot) to the default tab (slot) of the newly created window and afterwards gets rid of the dropped tab (slot).

For me, the whole package in its current form is already very usable and useful, but of course there is still a lot of potential for refinements and additional features.

If there is interest, I can create this PR in similar form based on master so that it merges cleanly. Please let me know.

feren:
GTK3 Widget Factory does have a button inside of a GTKNoteBook to the right, on Page 3 of the widget testing pages in it, so maybe looking into GTK3 Widget Factory's code to find out how they added a button to the Notebook might be something feasible to consider in order to program a similar feature into Nemo's GTKNotebook (tab bar) as a New Tab button?

PrasadGanesh:
hi,
if i want to contribute this project, whom should I ask and like where to start the discussion? is their any slack channel?

thanks and really sorry for asking here.

clefebvre:
I'm going to sound like a killjoy, but here goes. I don't see the need for this.

If the idea is to open a new tab before deciding where to go or without yet being able to access the destination, then why not simply use the existing toolbar. A new "New Tab" button, hidden by default (similar to the New Folder button), would do the trick.

I'll mark this PR as WIP for now since it's not ready in any case. It should be abandoned and closed though unless an actual need is identified. It doesn't fix any issue I can think of.

Jeremy7701:
A keyboard shortcut is available - CTRL+T
For two "tabs", the split pane design (F3) preserves more space for icons etc.

PR #2365, opened 2020-04-27
Nemo 4.4.2 Request : Time and Total-time in the sound files

JeSuisEnPanne:
how are you today ?

Nemo 4.4.2 Request :
Can you create useful information for audio files?
Information on the time of a sound file.
also, by selecting several files, is indicate the total time.

thank you. Bye


mtwebster:
Hi, there's a nemo plugin called 'nemo-media-columns' which makes additional columns available in the list view, showing media file attributes and metadata.

Unfortunately there's no way to have it give you a total in the selection, and it's not something easily accomplished given how extensions work.  The best that might be accomplished currently is writing some script that operates on the selected files list and displays a window (or terminal window) showing the results.  This could then be triggered from the context menu using nemo actions (see nemo preferences->plugins->Actions)

PR #2392, opened 2020-05-23
use debhelper 11 for integrated meson support

Fantu:
- using debhelper 11
- removed hacks for the minimal meson+ninja build on older debhelper 10


mtwebster:
This looks fine but is there any way to get my quiet build back? (where I don't see the command line for every compiled file)

Thanks

Fantu:
for the CI build fails is for xapps CI jobs named differently (used also for tags), already reported to clem in slack (see 2 week ago in history chat)

@mtwebster the flags are not changed, if this commit compile different from before for you I suppose is because with hacks for old debhelper build used custom build parameters on you device instead
about addition debhelper verbose FWIK there is DH_VERBOSE=1 (but is 0 by default if I remember good)

mtwebster:
Well, now it's back to normal.

PR #2535, opened 2020-10-16
Drop Debian menu file

Fantu:
From Pino Toscano patch for debian packages.
Drop Debian menu file (wrongly installed in gir1.2-nemo-3.0) since nemo already
provides a .desktop file.

### DIFF
--- a/debian/menu
+++ b/debian/menu
@@ -1,2 +1,0 @@
-?package(nemo):needs="X11" section="Applications/File Management" \
-  title="Nemo" command="/usr/bin/nemo" icon="/usr/share/pixmaps/nemo.xpm"


Fantu:
I not understand why CI still try lmde and mint and even if was changed in https://github.com/linuxmint/nemo/commit/08ed439e61ebb493ec5e50f0b60af12991b27fbc
@clefebvre @mtwebster any idea about this strange CI thing?

ItzSwirlz:
I'll be 100% honest, LM isn't up-to-date on packaging standards; they probably wouldn't care about this much.

mtwebster:
It's not about not caring, it's about time management. This isn't a top priority as long as it works properly, but we'll always accept improvements or accomodate changes for other desktops or distributions.

PR #2539, opened 2020-10-18
make sure pango development files are installed

Fantu:
Patch of Norbert Preining from debian packages.
With this check pango even if the version is lower that 1.44 (for example for
buster).
Added missed libpango1.0-dev in debian/control build-deps.

Note: I suppose Norbert set pango >=1.42 because is a version tested but probably an older version is supported, for example before meson port in configure.ac the min pango version required was 1.28.3, someone can confirm it or if a newer version is really required please?

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -20,6 +20,7 @@ Build-Depends:
  libgtk-3-dev (>= 3.9.10),
  libgtk-3-doc,
  libnotify-dev (>= 0.7.0),
+ libpango1.0-dev,
  libx11-dev,
  libxapp-dev (>= 1.4.0),
  libxext-dev,
--- a/meson.build
+++ b/meson.build
@@ -131,6 +131,9 @@ if libselinux_enabled
 endif
 conf.set('HAVE_SELINUX', libselinux_enabled)
 
+# make sure pango development files are installed
+pango = dependency('pango', version: '>=1.40.0')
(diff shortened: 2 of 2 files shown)


mtwebster:
Maybe 1.40 would be better.  Scanning this:
https://repology.org/project/pango/versions
That would cover a lot more, and I doubt anything before that matters.

Other than the 1.44 check there, I can't remember anything that would have required some recent minimum version.

Fantu:
@norbusan can you please tell me if you a searched on pango related code about the min version really needed or you did >=1.42 only because you tested on it?

norbusan:
@Fantu Hmm, I faintly remember that there was some back and forth concerning pango versions and abi breakage, but don't really remember the details.

Fantu:
I searched many (probably major) of pango function used in nemo and I found only pango_attr_insert_hyphens_new that is inside "ifdef HAVE_PANGO_144" that need pango >1.40
I also searched in nemo function added in 1.42 ( watching https://salsa.debian.org/gnome-team/pango/-/blob/debian/master/debian/libpango-1.0-0.symbols ) without found nothing, so I suppose is ok change "pango >=1.42" of this patch to "pango >=1.40" for more distro versions support

mtwebster:
Can you change it here or want me to after I merge?

PR #2547, opened 2020-11-01
Improve debian packaging

Fantu:
I merged some improvements from debian packages.
About debian/control from a fast search in history I not found where is added libxt-dev (here) and removed shared-mime-info (in debian https://salsa.debian.org/cinnamon-team/nemo) so I keep as it for now, I also keep selinux disabled. About these things someone can help me to understand what is better do (if changes is needed here or in debian packages) please?


mtwebster:
- Why dh_python3? The script in /usr/share/nemo/actions is just for a sample action, it's not really part of 'nemo' proper. (no big deal, just wondering).
- I don't see why libxt-dev needs to be kept. I don't see anything using it, and it's not brought into the meson build. I noticed ubuntu's nautilus package still has it as well, but that could just be they're as bad at maintaining their packaging as we are :smile: 
- We probably only require shared-mime-info in the package so the updater can run during installation. I don't see why we need it as a build dep.
- I can't speak on selinux, but keeping the existing behavior is best.

Fantu:
@mtwebster thanks for reply
if you don't want "debian: readd dh-python" exclude it (as there are only 2 "example files" there are python is not important to have it FWIK)
if also you don't found the utility of libxt-dev I suppose is ok remove it
if you don't found utility of shared-mime-info as build-dep probably can be safely removed (and keep it only in nemo deps) as debian packages did

PR #2551, opened 2020-11-09
Remove uft8 colon-like char in datetime strings

### LINKED ISSUE #2546: The colons (:) between HH:MM:SS become full-width (：) in Chinese UI
```
 * Nemo version (4.6.5)
 * Is issue with desktop or windowed nemo? (Windowed)
 * Distribution - (LMDE 4)
 * Graphics hardware *and* driver used (GTX1660s, 450.66)
 * 64 bit
 ```

**Issue**
图片
After the user's language change to Chinese, the colons become full-width. 

**Steps to reproduce**
Open Nemo settings, change the date form to the second, and it happens.

**Expected behaviour**
It should be like this.
图片
It's Thunar.

**Other information**
Thank you for solving this issue!

okaestne:
Fixes #2546
Introduced in 99102d4 as a design improvement.

This also fixes that you can copy the datetime string from a file's properties window intro other programs as a datetime string (e.g. python shell).

### DIFF
--- a/libnemo-private/nemo-file.c
+++ b/libnemo-private/nemo-file.c
@@ -4900,7 +4900,6 @@ nemo_file_get_date_as_string (NemoFile       *file,
 	gboolean use_24;
 	const gchar *format;
 	gchar *result;
-	gchar *result_with_ratio;
     int date_format_pref;
 
   	if (!nemo_file_get_date (file, date_type, &file_time_raw))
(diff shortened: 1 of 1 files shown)


mtwebster:
Looks like the utf-8 issue is fixed in glib now:
https://gitlab.gnome.org/GNOME/glib/-/commit/bc59892b1af2a8f4374a1ad2054ff9444151732b

PR #2557, opened 2020-11-13
Small improvements for libnemo-private

hduelme:
Removed not required counting variable.
Moved some variables, so that they are declared when needed.
Added documentation for nemo-file-operations.c.

### DIFF
--- a/libnemo-private/nemo-file-operations.c
+++ b/libnemo-private/nemo-file-operations.c
@@ -250,30 +250,33 @@ static char * query_fs_type (GFile *file,
 static int
 seconds_count_format_time_units (int seconds)
 {
-	int minutes;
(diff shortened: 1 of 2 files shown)


mtwebster:
Hi, a few things here:

- We'd prefer to keep the convention where durations are written as numeric expressions rather than just a single whole number - it makes it easier to read and understand how long it actually is, and far simpler if I wish to make a change (like changing 4 seconds to 5 seconds).
- I want to fix variable scopes (don't outside of the scope they're used), but we generally prefer to define them at the beginning of the scope, unless it's a very long function with unalike sections. That's not the case here - a couple of changes are appropriate (such as in free_xdg_dir_cache) but the others don't really add to readability.

PR #2559, opened 2020-11-14
nemo-mime-actions.c: Don't show mime button if the activation URI is null

### LINKED ISSUE #2442: Segfault when opening launcher pointing to invalid directory
```
 * Nemo 4.6.4
 * Desktop and Windowed
 * Mint 20
 * Virtualbox
 * 64 bit
 ```

**Issue**
Nemo segfaults when trying to open a .desktop file with `Type=Link` and `URL=` pointing to an invalid directory.

**Steps to reproduce**
Create `examples.desktop` and place the following content inside it

```
[Desktop Entry]
Version=1.0
Type=Link
Name=Examples
URL=file:///a/path/that/does/not/exist
```
[examples.desktop.gz](https://github.com/linuxmint/nemo/files/4874814/examples.desktop.gz)

Double click it with nemo.

**Expected behaviour**
Possibly throw an error and not segfault.

**Other information**
Originally discovered on nemo 4.0.6 on Ubuntu, but reproducible with a fresh Mint 20 live image in Virtualbox.
Backtrace with `thread apply all bt full`: [nemo_bt.txt](https://github.com/linuxmint/nemo/files/4874816/nemo_bt.txt)

icarter09:
This PR resolves #2442 

Added check for URI being null, which was a result when `nemo_file_get_uri` failed. Displays a simple message dialog to the user informing them an error occurred and avoids the previous problem of a segfault.

### DIFF
--- a/src/nemo-mime-actions.c
+++ b/src/nemo-mime-actions.c
@@ -1235,7 +1235,10 @@ application_unhandled_uri (ActivateParameters *parameters, char *uri)
 
     primary = _("Unknown file type");
     display_name = nemo_file_get_display_name (file);
-
+    if (display_name == NULL || g_strcmp0(display_name, "") == 0) {
+        g_free(display_name);
+        display_name = g_strdup(uri);
+    }
     if (enable_exec_button) {
         secondary =
             g_strdup_printf (_("The file \"%s\" has no known programs associated with it.  "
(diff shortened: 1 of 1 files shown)


mtwebster:
I can't make this new message appear - I addressed some of this already in my favorites commits - did you initially do this before favorites was merged? 

I get this:
image

Which isn't a crash, but also isn't helpful at all.

icarter09:
Yes. Clicking on the "Choose a program" would then create the segfault.

clefebvre:
I added two commits:

- one to fix the empty string in the dialog.
- one to tackle the issue differently and simply not show the "Choose a program" button if nemo_file_get_uri() returns NULL

In the original fix, I didn't like having to use another msgid and a succession of two open dialogs.

PR #2565, opened 2020-11-22
Fixes issue 942. Adds check for generic x applications

### LINKED ISSUE #942: Unknown type displayed with defined detailed type
In the list view, Type should never be "Unknown" whenever Detailed type is specified and displayed.

icarter09:
This PR resolves #942. 

Added entry for "application-x-generic" when checking for the given file's type.

### DIFF
--- a/libnemo-private/nemo-file.c
+++ b/libnemo-private/nemo-file.c
@@ -6862,10 +6862,6 @@ get_basic_type_for_mime_type (const char *mime_type)
         }
     }
 
-    if (basic_type == NULL) {
-        basic_type = g_strdup (_("Unknown"));
-    }
-
     g_free (icon_name);
 
     return basic_type;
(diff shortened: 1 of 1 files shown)


mtwebster:
Maybe it would be better to fall back to the g_content_type_get_description() if the basic type isn't known.  For instance an rdf file shows "RDF file" as the detailed type.

Check out: https://github.com/linuxmint/nemo/blob/d4980ff30bf994dea0a7b7ad1d25003fdf4639f3/libnemo-private/nemo-file.c#L6849-L6917

You could have get_basic_type_for_mimetype() return NULL instead of "Unknown", and use g_content_type_get_description() in get_description() if that happens.

icarter09:
Using the description looks much better now than defaulting to "Application".

PR #2599, opened 2020-12-22
Removes unused variables

icarter09:
Removes some variables that the build had detected as being unused or set, but not used.


ItzSwirlz (commented):
I'm a little concerned here, you sure this won't *impact* anything?

JosephMcc:
Concerned why? I really quick look shows these variables are never used.

collinss:
@ItzSwirlz Since this is c, it wouldn't even compile if the variables were used anywhere, so I'd say that passing ci is a pretty good indication it's safe. :)

ItzSwirlz:
Latest commit takes care of this, you can close this.

PR #2602, opened 2020-12-24
build: fix typelibdir in libnemo-extension/meson.build.

ericonr:
Without this commit, a build configured with --prefix=/usr and
--libdir=/usr/lib will generate typelibdir = "//lib/girepository-1.0",
which isn't correct.


ItzSwirlz:
Last time meson was updated (cinnamon repo), builds failed on s390x and arm64. I'll test

PR #2604, opened 2020-12-27
Add escaping for percent sign in action Exec

vilhalmer:
I ran into a situation where a tool I was trying to call as an action also uses `%` tokens in its format string. Nemo stops processing tokens if it hits one it doesn't know about, so there was no way to run this tool directly. I considered just making it ignore unknown token types, but decided it would be better to add `%%` to insert a literal `%` in the command, since you might need to pass one that _is_ a valid action token.


mtwebster:
Hi, can you add the PR description above to the commit message as well?  This adds some context (and keeps me from having to do a 'merge' commit to reference this pr).

Thanks!

vilhalmer:
Updated commit with a slightly polished description. :+1:

PR #2606, opened 2020-12-31
Increases emit time of change for progress update

### LINKED ISSUE #591: [Bug] File copy dialog "copying file..." don't match in the title and the window
nemo-file_copy

icarter09:
This PR resolves #591


mtwebster:
Issue # in commit!

PR #2609, opened 2021-01-01
Reorder Devices in sidebar

### LINKED ISSUE #561: partitions shown in reverse order
lmde nemo - partitions in reverse order

icarter09:
This PR resolves #561


mtwebster:
Would it be simpler to reverse the list from which we're adding devices?

See:
https://github.com/linuxmint/nemo/blob/master/src/nemo-places-sidebar.c#L914

Couldn't we just add: 
```
mounts = g_list_reverse (mounts);
```

icarter09:
Changed back the previous logic and added the g_list_reverse suggestion.

clefebvre:
The lists aren't ordered at all so reversing them doesn't help. Michael is implementing a new solution for this so I'm closing this PR.

PR #2641, opened 2021-01-17
Fixes issue with desktop sort not saving

### LINKED ISSUE #2239: Desktop item sort mode is not saved
```
 * Cinnamon version = 4.0.9
 * Distribution = Linux Mint 19.1
 * Graphics hardware = ACER TravelMate integrated graphics
 * 64 bit
 ```

**Issue**
When customizing the current monitor layout, if a sorting mode other than "name" is selected for items, that value is not maintained, because if you reopen the same menu, the value is always reset to "name"

**Steps to reproduce**
1. Right click on desktop
1. Choose "Customize"
1. Change "Sort items by" to "Size", "Type" or "Date"
1. Close window
1. Right click on desktop
1. Choose "Customize"
1. "Sort items by" is reset to "Name"

**Expected behaviour**
"Sort items by" should remember the previous value

icarter09:
This PR resolves #2239

### DIFF
--- a/src/nemo-desktop-overlay.c
+++ b/src/nemo-desktop-overlay.c
@@ -194,7 +194,7 @@ sync_controls (NemoDesktopOverlay *overlay,
         gtk_combo_box_set_active_id (GTK_COMBO_BOX (priv->direction_combo), combo_id);
 
         /* Sort type */
-        action = gtk_action_group_get_action (priv->action_group, "Vertical Layout");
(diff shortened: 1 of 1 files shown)


mtwebster:
Can you add more info to the commit message?

icarter09:
Updated the commit message.

mtwebster:
Can you put the issue # also in the commit message?

*runs away...*

icarter09:
Commit message updated to contain the issue #.

PR #2659, opened 2021-02-02
Fixes #2244. Checks for leading space when setting the location entry

### LINKED ISSUE #2244: Pasting text into the path box can append when it should overwrite
```
 * Nemo version (nemo --version) 4.2.3
 * Is issue with desktop or windowed nemo? Windowed
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) 19.2
 * Graphics hardware *and* driver used: Intel onboard; Mesa
 * 32 or 64 bit
 ```

**Issue**

If I paste text over a path in the nemo bar, then, if the text I am pasting starts with a space, the result is that the new text gets _appended_ to the old text.

**Steps to reproduce**

Have a path in the bar/box - e.g. '/etc/lightdm'. Have text starting with a space on the clipboard - e.g. ` /test` (_sic_). Paste over the existing text.

**Expected behaviour**

The selected text should be replaced, not appended to.

**Other information**

See screenshot of result.

I think this effects the alt-F2 box.

### PR DESCRIPTION
**Pull request by icarter09**
This PR resolves #2244

### DIFF
--- a/src/nemo-location-entry.c
+++ b/src/nemo-location-entry.c
@@ -330,11 +330,16 @@ nemo_location_entry_activate (GtkEntry *entry)
 	if (entry_text != NULL && *entry_text != '\0') {
 		uri_scheme = g_uri_parse_scheme (entry_text);
 
-		if (!g_path_is_absolute (entry_text) && uri_scheme == NULL && entry_text[0] != '~') {
-			/* Fix non absolute paths */
-			full_path = g_build_filename (loc_entry->details->current_directory, entry_text, NULL);
-			gtk_entry_set_text (entry, full_path);
-			g_free (full_path);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/nemo-location-entry.c [hunk 1]
```diff
@@ -330,7 +330,9 @@ nemo_location_entry_activate (GtkEntry *entry)
 	if (entry_text != NULL && *entry_text != '\0') {
 		uri_scheme = g_uri_parse_scheme (entry_text);
```
mtwebster, line 332: I think we should trim leading and trailing space before we actually handle it using something like `g_strstrip()` - then I think we can just let it be processed like before.

icarter09:
Added the logic to trim the leading/trailing space.

LinuxOnTheDesktop:
May I ask whether this fix has been released yet? For when accidentally I paste a path containing leading spaces into Nemo's bar, I get, still, a lot of undesirable oddness.

Issue #2624, opened 2021-01-09
Favorites ignore folder sorting order


kimimaru4000:
```
 * Nemo version 4.8.4
 * Is issue with desktop or windowed nemo? - Windowed
 * Distribution - Linux Mint 20.1 Ulyssa
 * GeForce RTX 2060 Mobile (Driver: nvidia-driver-450.102.04-0ubuntu0.20.04.1)
 * 64 bit
 ```
After favoriting an item, it always shows up in front of all other items. While this may be desirable for some, I'd like an option in Nemo to disable this behavior.

**Steps to reproduce**
1. Place at least 3 files ordered alphabetically in a folder.
2. Favorite a file later in the list.
3. Observe that the file is now bumped to the top.

Before favorite:
![NemoBeforeFavorite](https://user-images.githubusercontent.com/1388037/104104981-293de080-52a3-11eb-861e-211a0b8169e0.png)

After favorite:
![NemoAfterFavorite](https://user-images.githubusercontent.com/1388037/104105024-630ee700-52a3-11eb-997f-88af2722c35e.png)

This occurs in Icon, List, and Compact view.

**Expected behaviour**
The item is favorited and the sorting order remains the same.

**Other information**
I recommend adding an option to toggle this behavior so users can choose what they prefer.

PR #2662, opened 2021-02-06
Added option to toggle sorting favorites before other files

edaigle:
Took a crack at implementing an option to toggle sorting favourites before other files, as requested in #2624. 

Changes are based off of the existing "sort folders before files" option and use a new gsettings key:

`gsettings set org.nemo.preferences sort-favorites-first (true|false)`

Demo:

GIF Demo


mtwebster:
Looks good, thanks!

PR #2667, opened 2021-02-13
Issue 2666. Blocks emoji chooser from breaking Nemo

### LINKED ISSUE #2666: Ctrl + ; makes mouse disappear and get stuck on desktop or active window
```
 * Cinnamon version: 4.6.8
   - Please specify if you are using the daily builds PPA: No
 * Distribution: Mint "Ulyssa" 20.1
 * Sapphire Radeon HD 7950 Dual-X Boost, amdgpu
 * 64-bit
 ```
 * [xsession-errors.txt](https://github.com/linuxmint/cinnamon/files/5973755/xsession-errors.txt)

**Issue**

When accidentally pressing Ctrl + ; instead of Ctrl + L in nemo,
it would get the cursor stuck in a "always pressing left click" and unable to get out of it.

**Steps to reproduce**

1) Open up nemo or go to the desktop
2) Press Ctrl + ;
3) Now you're stuck there and your mouse disappears, being unable to switch to another window unless Ctrl + Tab is used.

**Expected behaviour**

Nothing would happen and Cinnamon would continue on as normal.

**Other information**
It prevents you from using any keybinds to restart Cinnamon, leaving you with only able to use the TTY to restart Cinnamon.
Even then, you would have to end up killing the program that Ctrl + ; was used on with `killall` to get your mouse working and showing again.

If that key combination was used on the desktop instead, you're unable to do anything regardless of restarting the cinnamon desktop entirely or not, and the only fix I found so far was restarting linux from the TTY.

It prevents you from properly typing onto some windows or even the cinnamon menu itself, only being able to type on one window even if another window is active. I haven't tested what other programs it would get stuck on since I don't want to risk restarting my computer again.

### PR DESCRIPTION
**Pull request by icarter09**
This PR resolves #2666

### DIFF
--- a/src/nemo-window.c
+++ b/src/nemo-window.c
@@ -1166,6 +1166,13 @@ nemo_window_key_press_event (GtkWidget *widget,
 	active_slot = nemo_window_get_active_slot (window);
 	view = active_slot->content_view;
 
+      /**
+       * Disable the GTK Emoji Chooser
+       */
+      if ((event->keyval == GDK_KEY_semicolon || event->keyval == GDK_KEY_period) && (event->state & GDK_CONTROL_MASK)) {
+          return FALSE;
+      }
+
 	if (view != NULL && nemo_view_get_is_renaming (view)) {
 		/* if we're renaming, just forward the event to the
 		 * focused widget and return. We don't want to process the window

### CODE REVIEW

File: src/nemo-window.c [hunk 1]
```diff
@@ -1166,6 +1167,13 @@ nemo_window_key_press_event (GtkWidget *widget,
 	active_slot = nemo_window_get_active_slot (window);
 	view = active_slot->content_view;
 
+      /**
+       * Disable the GTK Emoji Chooser
+       */
+      if ((event->keyval == GDK_KEY_semicolon || event->keyval == GDK_KEY_period) && (GDK_CONTROL_MASK + event->state == GTK_EMOJI_CHOOSER)) {
```
mtwebster, line 1173: event->state is a bitfield, so I think the second part of this condition should simply be:
```
(event->state & GDK_CONTROL_MASK)
```
Which checks is the control is being held, regardless of the state of other modifiers. This current patch doesn't work if you have capslock on or numlock off.

https://developer.gnome.org/gdk3/stable/gdk3-Windows.html#GdkModifierType

icarter09: Cleaned up the second part of the conditional.

PR #2668, opened 2021-02-14
Added option to change default password save settings in mount dialog.

chasep255:
I added added an option in the Edit/Preferences/Behavior menu to change the default selection when prompted for a password. 
 
Currently the default is hard coded to remember passwords for the duration of the session.  For security purposes I would prefer not to have the passwords for my encrypted drives being saved in the gnome keyring since any application running as myself can query it.  When prompted for the password I usually select the forget option, however sometimes I forget.  I would like to be able to change this default in the settings so I don't have to check that option every time.  

I added this change to all places I found calls to `g_mount_operation_set_password_save` except in `nemo-connect-server-dialog.c` since that file was using a different default.

### DIFF
--- a/gresources/nemo-file-management-properties.glade
+++ b/gresources/nemo-file-management-properties.glade
@@ -212,56 +212,6 @@ along with .  If not, see <http://www.gnu.org/licenses/>.
       </row>
     </data>
   </object>
-  <object class="GtkListStore" id="model64">
-    <columns>
-      <!-- column-name gchararray -->
-      <column type="gchararray"/>
-    </columns>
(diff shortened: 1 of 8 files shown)


clefebvre:
Hi,

This is a bit overkill, especially if it's visible in the UI prefs.

I'd like to suggest a better approach which would be simpler and more useful. Instead, we could keep a key in gsettings to remember the last selected option, and simply default to that. Basically the dialog would default to forget because that's what you selected the last time you were asked.

I'll close this PR for now. Feel free to create a new one if you like the suggested idea.

Issue #2674, opened 2021-02-18
[Feature Request] A way to open a Nemo tab from an external program


stevenxxiu:
```
 * Nemo version (nemo --version) nemo 4.8.4
 * Is issue with desktop or windowed nemo? windowed
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) Arch
 * Graphics hardware *and* driver used Nvidia proprietary driver
 * 32 or 64 bit 64 bit
 ```

**Issue**

I would really love to have a way to open a new tab with some path, in an existing instance of Nemo, by using some external command.

I frequently find myself opening Nemo from another application, say SmartGit or the terminal, and copy-pasting that path so I can open this as a tab in the existing instance of Nemo instead.

PR #2683, opened 2021-03-02
Added an `--existing-window` flag to open URLs in an existing window

stevenxxiu:
This implements the feature request https://github.com/linuxmint/nemo/issues/2674.

I implemented it by sending the `EXISTING_WINDOW` in the `options` variable to the `nemo_main_application_open()` function.

The function itself that opens new tabs is `open_tabs_in_existing_window()`, which finds an active instance of a Nemo window and opens new tabs there.

### DIFF
--- a/src/nemo-main-application.c
+++ b/src/nemo-main-application.c
@@ -66,6 +66,7 @@
 #include <pwd.h>
 #include <fcntl.h>
 #include <errno.h>
+#include <gdk/gdkx.h>
 #include <glib/gstdio.h>
 #include <glib/gi18n.h>
 #include <gio/gio.h>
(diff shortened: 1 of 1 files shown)


src/nemo-main-application.c
```diff
@@ -391,21 +392,73 @@ open_tabs (NemoMainApplication *application,
...
+                g_free(uri);
+            }
+
+            /* go to the last tab we opened */
+            NemoWindowPane *pane;
+
+            pane = nemo_window_get_active_pane (window);
+            nemo_notebook_set_current_page_relative (NEMO_NOTEBOOK (pane->notebook), n_files);
+
+            gtk_window_present (GTK_WINDOW (window));
```
mtwebster, line 431: can you use gtk_window_present_with_time instead?

Maybe like this:
```
// you'll have to add #include <gdk/gdkx.h> to the top of this file

gtk_window_present_with_time (GTK_WINDOW (window),
                              gdk_x11_get_server_time (gtk_widget_get_window (GTK_WIDGET (window))));

```
The reason being, window managers decide whether or not to actually give a window focus by the time of this call compared with their last 'user' event (like a click or keypress). Using gtk_window_present really calls _present_with_time but with a time of 0. The window manager will ignore this window's focus request and try to just mark it urgent instead (flashing in the window list for example).

If you instead provide a real time, the wm thinks it was the user asking for this window and it's ok to raise this window to the top. GdkEvents (from clicks and such) usually provide this time for us, but in this case, though this is probably caused by the user, we need to get the time ourselves.

This has been broken forever with Cinnamon - any window has been able to raise, so your current form works there. I recently fixed this, however, so right now on my cinnamon/muffin master, when I run with `--existing-window` it doesn't actually present the window, it just marks it urgent.

See:
https://github.com/linuxmint/muffin/commit/ee9a615d75cc27734400fb3ae54b923ec6995ee6

mtwebster:
Hi, this works well, just see the (long - sorry) inline comment.

Thanks

PR #2705, opened 2021-03-23
Nemo view order in supported image viewers

cerulean0:
Allow images to be viewed in supported image viewers in the same order as they are shown in the current nemo view (directory or search).

For example:

- If images are sorted in ascending date order, opening an image and pressing next/previous in the image view will go to the next/previous image chronologically.
- If the image is part of a search result, available images in the image viewer are limited to the other search results.

This is similar to how images are handled in Windows Explorer and Windows Photo Viewer, which I find so intuitive a feature, I didn't even think of it as one until moving to Linux on my personal machine.

So this is an implementation that seeks to address this in a way that doesn't require specific coupling between applications, or extensive changes to image viewers.

The interface the image viewer has to implement is a very simple one - it just has to accept a list of urls, and display them in the given order, rather than applying its own sorting. Bare bones image viewers like feh and sxiv already do this.

When opening a single image, nemo then checks if the default application to show it is on a list of supported ones, and sends the image viewer a list of all the image urls in the current view, in presented order. No extra application specific code / command line parameters are required.

Unfortunately the basic image viewers of various distros (xviewer, viewnior, ristretto) all seem to apply their own basic alphabetical sort to image lists, so there's an accompanying change to xviewer [here](https://github.com/linuxmint/xviewer/pull/134), since it's very likely this will be available on systems with nemo.

This is a change I think would add unavailable and useful functionality (particularly those moving from Windows; I've seen a few posts elsewhere about people missing this), and not get in the way for people who don't need it.

Passing big url lists around seemed acceptably performant from a few unscientific tests. It passed a folder of 4000 images to xviewer with no issue, which only thumbnailed images in its gallery preview as it needed them.

### DIFF
--- a/libnemo-private/nemo-global-preferences.h
+++ b/libnemo-private/nemo-global-preferences.h
@@ -74,6 +74,9 @@ typedef enum
 /* Activating executable text files */
 #define NEMO_PREFERENCES_EXECUTABLE_TEXT_ACTIVATION		"executable-text-activation"
 
+/* Image viewers to pass nemo view sort order to */
+#define NEMO_PREFERENCES_IMAGE_VIEWERS_WITH_EXTERNAL_SORT "image-viewers-with-external-sort"
+
 /* Spatial or browser mode */
 #define NEMO_PREFERENCES_ALWAYS_USE_BROWSER			"always-use-browser"
 #define NEMO_PREFERENCES_NEW_TAB_POSITION			"tabs-open-position"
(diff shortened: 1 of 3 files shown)


libnemo-private/nemo-global-preferences.h
```diff
@@ -74,6 +74,9 @@ typedef enum
 /* Activating executable text files */
 #define NEMO_PREFERENCES_EXECUTABLE_TEXT_ACTIVATION		"executable-text-activation"
 
+/* Image viewers to pass nemo view sort order to */
+#define NEMO_PREFERENCES_IMAGE_VIEWERS_WITH_EXTERNAL_SORT "image-viewers-with-external-sort"
```
cerulean0, line 78: It seemed useful to have this controllable by property, but felt a bit low level to be referencing executable on the preferences page, so it's a hidden property for now. This could be supplemented with an extra toggle property to enable it for any in this list.

mtwebster: This works - we can add more as they're discovered.

I don't see any reason to be able enable/disable this behavior - it's ok as the 'normal' behavior I think. It doesn't interfere with the single selection behavior (browsing the entire folder).

PR #2721, opened 2021-04-04
bookmark-list: Clean up and close memory leak

ItzSwirlz:
Clean up do_finalize method. Use NemoBookmarkList instead
of casting itself every line.

Also, properly free the GFileMonitor when it isn't neccessary.
g_file_monitor_cancel() doesn't free the object.

Taken from:
https://github.com/GNOME/nautilus/commit/ec4241a80b029d0bb13d3f95a963d5f7bd5acd1a
https://github.com/GNOME/nautilus/commit/179e0bfaddcf29d8a7f8074cc924d5aa45b1a954

### DIFF
--- a/src/nemo-bookmark-list.c
+++ b/src/nemo-bookmark-list.c
@@ -238,16 +238,19 @@ clear_bookmarks (NemoBookmarkList *bookmarks, GList *list)
 static void
 do_finalize (GObject *object)
 {
-	if (NEMO_BOOKMARK_LIST (object)->monitor != NULL) {
-		g_file_monitor_cancel (NEMO_BOOKMARK_LIST (object)->monitor);
-		NEMO_BOOKMARK_LIST (object)->monitor = NULL;
+	NemoBookmarkList *self = NEMO_BOOKMARK_LIST (object);
+
+	if (self->monitor != NULL)
+	{
+		g_file_monitor_cancel (self->monitor);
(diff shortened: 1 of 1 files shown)


JosephMcc:
Why did you only clean up part of the casts in that function and not the others?

ItzSwirlz:
> Why did you only clean up part of the casts in that function and not the others?

I only try to make small patches. It helps maintainers if they want to make patches but whatever.. it was also where since I'm still learning API with Gtk/Nemo I wasn't 100% confident.

src/nemo-bookmark-list.c
```diff
@@ -238,16 +238,19 @@ clear_bookmarks (NemoBookmarkList *bookmarks, GList *list)
...
 	}
 
-	g_queue_free (NEMO_BOOKMARK_LIST (object)->pending_ops);
+	g_queue_free (self->pending_ops);
 
-	clear_bookmarks (NEMO_BOOKMARK_LIST (object), NEMO_BOOKMARK_LIST (object)->list);
+	clear_bookmarks (self, self->list);
 
-    g_object_unref (NEMO_BOOKMARK_LIST (object)->volume_monitor);
+        g_object_unref (&self->volume_monitor);
```
mtwebster, line 253: ditch the & operator - that's only required for g_clear_object() so it can set the pointer to NULL for you.

PR #2744, opened 2021-05-02
Made some indentation corrections

myaashish:
While going through code to understand its working, I came across some part of code where indentation was not looking correct, made changes as per my understanding about correct indentation practices. Please do correct me if my changes are unexpected or wrong.
Thank you

### DIFF
--- a/libnemo-private/nemo-icon-container.c
+++ b/libnemo-private/nemo-icon-container.c
@@ -526,8 +526,6 @@ icon_get_row_and_column_bounds (NemoIconContainer *container,
 			bounds->y1 = MAX (bounds->y1, one_bounds.y1);
 		}
 	}
-
-
 }
 
 static void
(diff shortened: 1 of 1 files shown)


mtwebster:
Hi,

We try to avoid making whitespace-only changes like this.

We like to use spaces instead of tabs. Nemo (and many other programs) originally used tabs.

You've converted the lines you've changed to spaces, but the surrounding lines are mostly tabs, so the code still looks odd depending on how you view it (look at the diff here on github and compare it to what you see in an editor).

Instead, we try to fix whitespace/formatting when we make substantive changes instead, and fix blocks of code (or entire functions) as appropriate to make sure they're readable, rather than just the lines we change.

Another issue is that if I view a 'git blame' of a file, I would see this commit for a given line you've changed here, rather than the commit that actually introduced the code on that line (take a look at https://github.com/linuxmint/nemo/blame/1408eccdd100fe67850d31f27a02216f4e2472d5/libnemo-private/nemo-icon-container.c#L646-L655) - this is not a huge deal, but can be annoying when searching thru old commits.

I appreciate the effort and am not trying to be discouraging, but I wouldn't worry much about whitespace issues, except when you're modifying actual code.

Thanks

PR #2747, opened 2021-05-13
Fix #2347 - Nemo menu unaccessible after hiding it

### LINKED ISSUE #2347: Nemo menu unaccessible after hiding it
```
 * Nemo version (nemo --version)
4.4.2

 * Is issue with desktop or windowed nemo?
Windowed

 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...)
Arch
 * Graphics hardware *and* driver used
Intel
 * 32 or 64 bit
64
 ```

**Issue**
After hiding the menu pane, the menu is not accessible anymore. Even though I can make it visible (by rightclicking or pressing "Alt"-key) it dissapears instantly the moment I try to open any of the menus (file, edit, view, go, ...).

**Steps to reproduce**

Reopen nemo. Try again...

**Expected behaviour**
The menu should stay open I guess...

**Other information**
Reseting nemo by renaming ~/.config/nemo doesn't work either

kysrpex:
The menu bar seems to lose focus after a sub-menu appears. As a workaround, one may check whether some element is selected before hiding it.

I added also some more code (I am no C programmer, this part can be improved for sure) to hide the menu immediately after choosing an option, otherwise there is still a small delay and the experience feels weird.

~~**Remark**: on a touchscreen, you can't keep your finger pressed against the options and then release to select, so if you do not hit the proper option on your first try, then the menu also closes. This is sub-optimal.~~

~~**Remark 2**: this also seems to happen with a mouse.~~

### DIFF
--- a/src/nemo-window.c
+++ b/src/nemo-window.c
@@ -563,7 +563,7 @@ hide_menu_on_delay (NemoWindow *window)
 }
 
 static gboolean
-on_menu_focus_out (GtkWidget *widget,
+on_menu_focus_out (GtkMenuShell *widget,
                    GdkEvent  *event,
                    gpointer   user_data)
 {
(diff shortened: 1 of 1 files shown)


j-james:
This fix works. I do also see what you mention in your remark, though.

mtwebster:
Looks like it doesn't affect anything under x11, though I don't quite get what you mean with your remarks, can you be a bit more descriptive? I don't notice any issue with selection when holding the mouse down after activating the menu.

kysrpex:
> Looks like it doesn't affect anything under x11, though I don't quite get what you mean with your remarks, can you be a bit more descriptive? I don't notice any issue with selection when holding the mouse down after activating the menu.

I just realized that it is unrelated, it also happens without the patch, without auto-hiding the menu, and even on Nautilus, Geary or other apps. Firefox gets it right though. In any case this would be a feature request for an upstream project (Gtk?). You can forget about those remarks, sorry :sweat_smile:.

PR #2750, opened 2021-05-23
Implement content search

mtwebster:
Improve search, implement content search.

- Allow showing number of matches in each file as a column in the results.
- Shows search result snipped in file tooltip.
- Support regex searches.
- Turned search bar into a glade file so it can be modified more easily.
- Folders (either absolute path or folder name -- .git for example) can be excluded via a gsettings list.

Currently, text/plain content type is supported natively. There is a mechanism to support additional mimetypes with external programs.

Currently I've added support for:
 - pdf/ps using ps2txt, pdf2txt or pdftotext.
 - open office using odt2txt.
 - image exif data using the exif tool.
 - mpeg id3 data using the id3 tool.

Additional ones can be added by users. Their task is to output whatever content deemed worthy of content search to their stdout (this doesn't even need to be the true contents of the file, simply what you send to stdout). These aren't hard dependencies - if the programs aren't there, you just won't have support for that content type.

There are some internal settings to remember UI toggle positions, and exposes a couple of free options for changing how GRegex behaves.

todo:
- ~~a history feature for the search entries (to replace existing 'saved search' features.~~
- ~~pdf2txt is crazy slow - maybe a better pdf extractor if one exists, or create one.~~ - added support for pdftotext
- ~~Add search helper programs to apt recommends?~~
- ~~Add search hit snippet to the tooltips of results.~~
- ~~Indicate when there are no results.~~
- ~~show content search as disabled for non-native filesystems~~
- ~~ms doc converter~~


Jeremy7701:
I'd suggest `pdftotext` as a much faster converter than `pdf2txt` - it is part of the Debian `poppler-utils` package.

mtwebster:
@Jeremy7701 Thanks for that suggestion, night and day faster.

PR #2753, opened 2021-05-31
make ps2txt search helper work on non-debianlike distros

eli-schwartz:
Specifically, there is no such thing as ps2txt and AFAICT never has been. But debian has for some decades been providing a convenience symlink from ps2ascii so that you can invoke it as 'ps2txt' as well, resulting in much confusion from people who actually try to do so, start relying on doing so, and then get bug reports for their software from other distros where this just plain does not work.

### DIFF
--- a/files/usr/share/nemo/search-helpers/ps2ascii.nemo_search_helper
+++ b/files/usr/share/nemo/search-helpers/ps2ascii.nemo_search_helper
@@ -1,5 +1,5 @@
 [Nemo Search Cat Helper]
-TryExec=ps2txt
-Exec=ps2txt %s
+TryExec=ps2ascii
+Exec=ps2ascii %s
 MimeType=application/ps;
 Priority=100;


Jeremy7701:
According to the man page of ps2txt -> ps2ascii, it is "ps2ascii - Ghostscript translator from PostScript or PDF to ASCII".
The man page claims it's not great and recommends the use of `pstotext`

However, Debian said (in 2018) that due to there being no updates since 2004, that pstotext didn't even make it into Debian 10. ;)
So unless one of the many `ps -> pdf -> text` options should prove "better", the `ps -> text` option isn't going to be 100%.

eli-schwartz:
I'm not here to make a judgment call about which ps converter is best.

I'm here to make a judgment call about whether nemo should look for this one under its actual name, or an imaginary fantasy name invented by some Debian maintainer for undocumented reasons. The only part of the Debian rationale I care about is "why was this frustrating name alias ever made, and did you think about the pain you were going to cause to other linux distributions?"

Jeremy7701:
There isn't a real choice.
You can use the symbolic link and/or you can use the "real" name.
Since its called `ps2ascii` I'm sure it isn't going to work well for most non-English users.

eli-schwartz:
I'm sorry, but at this point I'm totally lost. What do your comments have to do with this PR?

It seems like you don't have anything that's on topic to say about my proposed change. So please don't say anything.

If you feel your words are important to say -- please open a new issue to talk about it, rather than dragging my PR into it. Thanks.

leigh123linux:
@Jeremy7701  I'm locking this as I'm sick of getting spammed with worthless comments.

mtwebster:
Let's make sure things are kept civil please.

Issue #2352, opened 2020-04-13
(Feature request) "Keep both files" file conflict handling option


Minabsapi:
**Nemo version: 4.2.3**

While I was cleaning and moving some files I discovered there were no handling option to keep both files in Nemo when a file name conflict happened. The only options provided are replacing the file, renaming it manually **everytime**, ignoring (which I wonder the use), or cancel the operation.

Is it possible to add such option, handling the case on every conflict with the "Apply this action for all files" by automatically renaming the file with a pattern such as what can be found with folders: "conflictingfilename (1).txt" ?
It would really make life easier, and I'm really surprised it hasn't been implemented yet

www-MKRD-info:
I WILL PAY $ TO ADD SUCH AUTOMATIC RENAMING, WHICH IS LITERALLY PRESENT ELSEWHERE IN LINUX (FIREFOX DOWNLOADING DUPLICATES, MANUAL DUPLICATION ALTHOUGH I AM NOT IN FAVOR OF "ANOTHER COPY"), AND SAME AS PRESENT IN WINBLOWS.
TO BE CLEAR, I WANT (2), (3), (4), AND NOT "A COPY", "ANOTHER COPY", ETC.
I DO A LOT OF FILE OPERATIONS. THIS WILL IMPROVE MY LIFE.
WHILE WE ARE ON THIS TOPIC, MAYBE ADD AN OPTION INSTEAD OF NUMERIC INCREMENTING TAG, USE A DATE AND/OR TIMESTAMP INSTEAD!!! SAME AS WHAT FREEFILESYNC DOES ON MY SYSTEM.

icarter09:
@www-MKRD-info we appreciate the feedback, but there's no reason be commenting in all caps or adding in extra non-needed information. The team is working hard to implement not only your request, but many others. We welcome any and all pull requests so please feel free to implement the feature request as well. Thanks.

www-MKRD-info:
Changed to the way you like. Wasted tens of seconds, over doing this many, many times times over time, add up in my life. Been waiting for this to be implemented for years. Hence the language.

PR #2841, opened 2021-10-03
Added new "Keep both files" conflict handling option

Minabsapi:
## Description
This feature appears in the GUI as a new button "Autorename" added to all the different conflict dialogs
This uses the already implemented "unique name" policy that allows copying/moving a file/folder with the same name than another in destination folder by renaming it automatically in this fashion:
"example file (copy).txt" "example file (another copy).txt" "example file (3rd copy).txt" etc.
The main interest of this new feature is its cumulation with the "apply to all" checkbox, allowing automatic renaming for all conflicts.

## Bugs
No bug were found upon testing all relevant use cases

## Feature notes
Only the functioning feature was implemented. Translation is left to the maintainers / other contributors.

## Pull request notes
This addition is originally intended to be merged to the maintenance branch of the 4.4 version, as it was the only version I could personally test on my Ubuntu 20.04 machine; however it shouldn't conflict with any other version and is strongly encouraged to be merged to the latest versions, as it's a logical feature that is showcased in any classical file browser e.g. Windows Explorer
This closes the issue #2352, as well as some others treating of this subject that may or may not have been mentioning the issue I just linked.

## General notes
I profit of this merge pull to make a small complaint: such a small feature with a very few lines of code took a considerable time (around a week of evenings) to implement because of the lack of clear documentation of this repository. Despite the code being, I admit it, very clear and concise, I'd humbly ask some maintainers to write at least a small documentation that would help new contributors find themselves in this codebase

### DIFF
--- a/libnemo-private/nemo-file-conflict-dialog.c
+++ b/libnemo-private/nemo-file-conflict-dialog.c
@@ -57,6 +57,7 @@ struct _NemoFileConflictDialogDetails
 	GtkWidget *checkbox;
 	GtkWidget *rename_button;
 	GtkWidget *replace_button;
+	GtkWidget *auto_rename_button;
 	GtkWidget *dest_image;
 	GtkWidget *src_image;
 };
@@ -546,6 +547,12 @@ nemo_file_conflict_dialog_init (NemoFileConflictDialog *fcd)
 				_("_Skip"),
 				CONFLICT_RESPONSE_SKIP,
 				NULL);
+
+	details->auto_rename_button =
+		gtk_dialog_add_button (dialog,
+					   _("Auto_rename"),
+					   CONFLICT_RESPONSE_AUTO_RENAME);
+
 	details->rename_button =
 		gtk_dialog_add_button (dialog,
 				       _("Re_name"),
--- a/libnemo-private/nemo-file-conflict-dialog.h
+++ b/libnemo-private/nemo-file-conflict-dialog.h
@@ -62,8 +62,9 @@ struct _NemoFileConflictDialogClass {
 enum
 {
 	CONFLICT_RESPONSE_SKIP = 1,
-	CONFLICT_RESPONSE_REPLACE = 2,
-	CONFLICT_RESPONSE_RENAME = 3,
+	CONFLICT_RESPONSE_AUTO_RENAME = 2,
+	CONFLICT_RESPONSE_REPLACE = 3,
+	CONFLICT_RESPONSE_RENAME = 4
 };
 
 GType nemo_file_conflict_dialog_get_type (void) G_GNUC_CONST;
--- a/libnemo-private/nemo-file-operations.c
+++ b/libnemo-private/nemo-file-operations.c
@@ -98,6 +98,7 @@ typedef struct {
 	gboolean skip_all_error;
 	gboolean skip_all_conflict;
 	gboolean merge_all;
+	gboolean auto_rename_all;
 	gboolean replace_all;
 	gboolean delete_all;
 } CommonJob;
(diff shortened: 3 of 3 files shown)


Jeremy7701:
1. Would the many current cases where autorename occurs silently, such as 'duplicate' or 'create new file' or 'copy and paste in the same directory' etc be affected?

2. Just because a file is copied into a directory with a matching name doesn't **always** mean that the two files are 'copies' in any sense. I'd prefer to see some human oversight (as at present).

Minabsapi:
> 1. Would the many current cases where autorename occurs silently, such as 'duplicate' or 'create new file' or 'copy and paste in the same directory' etc be affected?

They aren't affected as it isn't related to the new introduced feature which acts only in case of conflict; nemo's current handling of these situations are kept.

> 2. Just because a file is copied into a directory with a matching name doesn't always mean that the two files are 'copies' in any sense. I'd prefer to see some human oversight (as at present).

I agree; know however that it was intended to be somewhat simplistic. Most people needing such feature don't really pay attention to the name of the file immediately, they just need it copied / moved for later renaming or even not.
Also, the "(copy)", "(another copy)" etc. mechanism used here was already existing and hasn't been introduced by this commit, simply reused. I don't think it's the best naming convention either and would prefer a more straightforward such as "(1)", "(2)" etc. But this isn't related here and such change should be discussed and contributed in another issue / pull request. (Note: I saw a nemo setting yesterday that enables you to set another naming pattern. Could it help with this?)

Jeremy7701:
Since this doesn't force a behaviour, but is simply an option, I don't have strong feelings against it.
However, it makes it too easy for a new user to "lose files".

It's obviously up to the Mint/LMDE teams to decide if they want to implement your change.

You might be interested in the new `bulky` add-on in `nemo 5.0.3` (the current stable version) which allows bulk renaming of files. This could be used either in the source or target directories to rename potential conflicts before doing any copy/move operations.

Minabsapi:
No behavior is forced indeed; nothing is changed, this is just intended to be a handy addition.
I don't quite get how it makes it easy to "lose" files as you said. As an anecdote, the only way I lost files in such configuration is when the feature wasn't implemented and I had to use the command line to do the job, and as it was late in the night and I'm pretty (not) at ease with Bash it turned out to be a catastrophe.

Thank you for the suggestion, unfortunately Nemo cannot be updated to this version on my distribution as already said. But I'll keep that in mind for the next stable distribution upgrade.

Jeremy7701:
It's very easy to "lose" files.

Suppose I upload my recent holiday snaps into my pics archive and (to save time) I take the new option.
In 6 months time when I come to look for a particular file, I will have forgotten that it might be listed as a '(copy)' of something else instead of a meaningful name and so appears to have disappeared.

Minabsapi:
It's all about you in this kind of case. There's nothing more to say

clefebvre:
I like it and it works well, and it's useful because nemo doesn't show these issues until we're halfway in the file operation.. so you don't have the opportunity to solve the issue before, but this gives the opportunity to do it after, without looking at each file in detail while effectively have the operation paused.

I agree with @Minabsapi also that this prevents losing files and not the opposite. You might delete "foo (copy)" thinking it's the same as "foo", but you don't have to. Without this feature, only one foo remains, the other one is either gone or in a different place altogether, full of other files which are strictly identical.

The one thing I don't like is the button label. We can't have something like that in a dialog and there's no room for a full sentence. I'm going to use "Duplicate" for a lack of a better word.

clefebvre:
I forgot to mention "Rename" is already used in the "Overwrite" button, when the entry is used to select a new name.

clefebvre:
damn it, sorry... I didn't notice this was against 4.4-maintenance, I'll clean it up asap.

clefebvre:
Merged on master at https://github.com/linuxmint/nemo/commit/a865bb2b03c70ee3ac685b38fe03e7508cff1bfc.

Note: I tried to find that feature in Windows 10 but I couldn't find it. Windows shows the conflicts before starting the operation, which is pretty cool though.

Minabsapi:
Thank you a lot for considering this. Greatly appreciated!

PR #2864, opened 2021-11-07
epub2text.sh

Jeremy7701:
You may/will want to alter this; html2text has problems displaying unusual characters.
On the Plus side its surprisingly fast - a search of 500 epubs took about 30 seconds.

### DIFF
--- a/search-helpers/home/jeremy/.local/share/nemo/search-helpers/epub2text.sh
+++ b/search-helpers/home/jeremy/.local/share/nemo/search-helpers/epub2text.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+# Requires unzip package
+# ==================================
+# 1. Get a list of all xhtml/html/htm files, exclude titlepage.xhtml (if present)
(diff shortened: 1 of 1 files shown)


Fantu:
use meson to install and make it executable and put in same PR with the other part (https://github.com/linuxmint/nemo/pull/2865) with correct path

Jeremy7701:
> use meson to install and make it executable and put in same PR with the other part (#2865) with correct path, probably is better also rename it to nemo-epub2text is good add also html2text to nemo depends in debian/control

Thanks Fantu,
As you can guess I've never used GIT etc before; I've managed to merge #2864 & #2865 into a #2867.
I've renamed epub2text.sh to nemo-epub2text, as you suggested.
However, I don't know how to make it executable via meson.
I don't understand the part about "html2text to nemo depends in debian/control" since html2text is not a dependency of any of the standard Debian packages.

Jeremy

clefebvre:
Replaced with https://github.com/linuxmint/nemo/pull/2867

PR #2865, opened 2021-11-07
epub2text.nemo_search_helper

Jeremy7701:
I could only get this to work with an absolute path.


Fantu:
this point to your home (/home/jeremy/...) and file to execute is missed in PR so is wrong and incomplete

clefebvre:
Replaced with https://github.com/linuxmint/nemo/pull/2867

PR #2885, opened 2021-12-08
search: add a PDF search helper

marcusmueller:
Based on the pdftotext poppler utility.

For historical context: This arose from https://unix.stackexchange.com/questions/680655/is-it-possible-to-integrate-pdfgrep-into-nemo-search/680659 and me wanting that for a long time, too.

Signed-off-by: Marcus Müller <marcus@hostalia.de>

### DIFF
--- a/search-helpers/application-pdf.nemo_search_helper
+++ b/search-helpers/application-pdf.nemo_search_helper
@@ -1,0 +1,5 @@
+[Nemo Search Helper]
+TryExec=pdftotext
+Exec=pdftotext %s
+MimeType=application/pdf;
+Priority=100


mtwebster:
I don't understand - we already have this: https://github.com/linuxmint/nemo/blob/master/search-helpers/third-party/pdftotext.nemo_search_helper

marcusmueller:
yeah, for some reason I skipped the third-party subdir, sorry for the noise.

mtwebster:
No problem, but it made me check to make sure it was working :)

marcusmueller:
@mtwebster Happy you're not angry :) Anyway, I realized that most users will probably want to be able to search PDFs, so adding poppler-utils to the debian/control Suggests: list seemed wise. Consequentially, just added all third-party search helper dependencies: #2886.

PR #2886, opened 2021-12-08
debian/control suggest third-party search helper dependencies

marcusmueller:
This was just going through all the third-party search plugins,

and looking for TryExec:, then checking for packages.

Intentionally omits 'ghostscript', as that can be heavyweight.


mtwebster:
Hi, these are already listed as dependencies.  They're promised functionality for nemo in Mint, so we require them.

PR #2913, opened 2022-01-21
Add option to disable or enable content type detection of mounted media.

patlefort:
Reasons to add this option: I don't need that functionality, it is wasted cycles and it takes too much space away from the file list.

I added the option in the media handling section, although it is a nemo specific option but is very much a media handling option, is it ok?


mtwebster:
Thanks - nemo is probably best, as it's where the related options live. There's sort of overlap with removable media options in Cinnamon but that's another issue.

PR #2924, opened 2022-02-05
Add "Copy as path" item to the menus

QuickActuator64:
This makes it very easy to quickly copy the path of a file e.g. so it can be pasted into another program that takes a path. Instead of copying the path from the pathbar and manually adding the name of the file, it now only takes two clicks to get the full path of one or more files.
The item can be found in the right-click context menu as well as the Edit menu from the MenuBar. The items are insensitive when nothing is selected, just as the "Copy" item. A gscheme entry has also been added so users can hide it (shown by default though).

Tested:
* Builds & runs successful
* Show/Hide settings works
* Copying location of single & multiple files works
* Copying location of local & remote (tested with smb) files works

The code changes should be following the [`nemo style guide`](https://github.com/linuxmint/nemo/blob/master/docs/style-guide.html).


smurphos:
Is this needed? Standard copy does a copy as path if the paste destination is a terminal/window or text editor already.

QuickActuator64:
I didn't know that, but I have tested it and yeah this feature seem to be obsolete. I will close this pull request then. At least I have learned something new...

PR #2952, opened 2022-03-29
Add !NEMO_IS_DESKTOP_WINDOW() guards to callbacks

### LINKED ISSUE #2921: Pressing ctrl+d when desktop is focused creates invalid bookmark
* Nemo version 5.2.4
 * Is issue with desktop or windowed nemo? desktop
 * Distribution - Ubuntu 22.04

**Issue**

Pressing ctrl+d (the keyboard shortcut to create a bookmark) when the desktop is focused creates an invalid bookmark. When attempting to browse to this bookmark in nemo, an error is displayed with the following two lines:

Could not display "x-nemo-desktop:///".
Nemo cannot handle "x-nemo-desktop" locations.

Of note, I am not actually trying to create a bookmark to the desktop.  Rather, since in a terminal window, ctrl+d sends EOF, closing the terminal when the terminal's initial shell exits, when I overestimate how many nested shells or SSH sessions I'm in and press ctrl+d one too many times, the terminal exits, and the last ctrl+d is sent to the desktop.  Thus, we may want to consider removing the feature of ctrl+d creating a desktop bookmark if we think that it is only ever accidentally triggered.

**Steps to reproduce**

Press ctrl+d when desktop is in focus.  Browse to newly created bookmark in nemo.

**Expected behaviour**

Either ctrl+d should create a working bookmark to the desktop or ctrl+d should have no effect.

**Other information**

Another user may have experienced this issue in 2014:
https://forums.linuxmint.com/viewtopic.php?t=173971

jknockel:
Fixes #2921.

Some action callbacks are already guarded with !NEMO_IS_DESKTOP_WINDOW() checks, but not enough unfortunately.  Guards are added to these callbacks for the following justifications:

- CTRL+Q: action_close_all_windows_callback (segfaults)
- CTRL+L: action_menu_edit_location_callback (does nothing but print nonfatal assertion failures)
- CTRL+D: action_add_bookmark_callback (adds bogus bookmark)
- CTRL+B: action_edit_bookmarks_callback (crashes nemo-desktop with fatal assertion failure)
- ALT+P: action_plugins_callback (fails to load extensions)
- F9: action_show_hide_sidebar_callback/nemo_window_get_sidebar_id (silently changes global sidebar preference even though desktop has no sidebar)
- / or ~: nemo_window_prompt_for_location (does nothing but print nonfatal assertion failures)

### DIFF
--- a/src/nemo-window-menus.c
+++ b/src/nemo-window-menus.c
@@ -307,9 +307,11 @@ action_plugins_callback (GtkAction *action,
 {
     GtkWindow *window;
 
-    window = GTK_WINDOW (user_data);
+    if (!NEMO_IS_DESKTOP_WINDOW (user_data)) {
+        window = GTK_WINDOW (user_data);
 
-    nemo_file_management_properties_dialog_show (window, "plugins");
+        nemo_file_management_properties_dialog_show (window, "plugins");
+    }
 }
 
 static void
@@ -525,7 +527,9 @@ static void
 action_close_all_windows_callback (GtkAction *action,
 				   gpointer user_data)
 {
-	nemo_application_close_all_windows (nemo_application_get_singleton ());
+	if (!NEMO_IS_DESKTOP_WINDOW (user_data)) {
+		nemo_application_close_all_windows (nemo_application_get_singleton ());
+	}
 }
 
 static void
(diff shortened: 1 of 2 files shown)


jknockel:
Gah, I just noticed a regression.  I added [WIP] until I fix it...

jknockel:
OK just a typo thankfully. Sorry for the noise. This is ready again for review. :)

Issue #2931, opened 2022-02-17
Replace libnotify with GNotification


WhyNotHugo:
`libnotify` is pretty much unmaintained at this point.

[GNotification](https://wiki.gnome.org/HowDoI/GNotification) is the way to go now, and should not introduce any new dependencies.

These commits from Nautilus should provide good reference:

- https://gitlab.gnome.org/GNOME/nautilus/-/commit/3c7ef4ae825308ec723eb684a88baf30348a9867
- https://gitlab.gnome.org/GNOME/nautilus/-/commit/a98c9300f4f6eedba8f6a54411e682107f259eda
- https://gitlab.gnome.org/GNOME/nautilus/-/commit/53269b41d358fa30806a993a400ca0c147d0cd10

I'm willing to give this a shot if the proposal makes sense to the maintainers.

PR #2956, opened 2022-04-02
Port to GNotification

okaestne:
Replace libnotify with gio's GNotification and thus remove a dependency.

Fixes: #2931

I tested the unmount pending/done notifications, but didn't manage to get the "All file operations have been successfully completed" notification though. Wasn't able to trigger that one somehow.


WhyNotHugo:
Nice, thanks!

Issue #2708, opened 2021-03-24
Feature request: Add native exfat  character support.


eyalroz:
* Nemo 4.8.6 (built from source, using xapp 2.0.7 and cinnamon-desktop 4.8.1 built from source)
* Windowed (but it doesn't matter)
* Distribution: Devuan Beowulf
* Graphics hardware: Irrelevant
* 64 bit

When I copy a file from a filesystem which supports some set of characters, to a filesystem which supports a different set, and the filename has characters in the difference of the sets - I get an error dialog, which tells me "No such file or directory" about the target filename.

That's the wrong error; and you have to know what's wrong in order to decipher it.

To reproduce:

1. Have an ext4 filesystem mounted
2. Have an exFAT filesystem on a USB-connected HDD, mounted via fuse
3. Create a file named "hello:world" somewhere on the ext4 filesystem
4. Using nemo, try copying this file to a writable location on the mounted filesystem. 

The kernel version may also be playing a part here, so if you're not on the exact same distro as me, try at least to be using Linux 5.9.0.

This should get you the invalid/misleading dialog mentioned above.

leigh123linux:
nemo-3.8.x is EOL, feel free to reopen if you can reproduce on a supported version.

eyalroz:
@leigh123linux : I can't reopen myself, only project members can. please reopen.

leigh123linux:
You will need to provide more info, please provide full steps to reproduce.

eyalroz:
@leigh123linux : See edit, but it's not as though I'm talking about some esoteric scenario.

leigh123linux:
I can't reproduce to issue, hello:world is automatically renamed to hello_world when copying from ext4 to fat.

https://github.com/linuxmint/nemo/blob/master/libnemo-private/nemo-file-operations.c#L3392

eyalroz:
@leigh123linux : Interesting! I've constricted the reproduction instructions.

Also, the automatic-renaming is another reason why this bug and #2709 are different :-)

leigh123linux:
I can reproduce the issue with native exfat, it looks like glib will need exfat fix 

https://gitlab.gnome.org/GNOME/nautilus/-/merge_requests/565#note_847645

https://github.com/GNOME/glib/blob/46097e9b8e4ad253e0e3320e929cc0fa3df84d5d/gio/glocalfile.c#L563

eyalroz:
@leigh123linux : 

[...]

PR #2967, opened 2022-04-15
Add native exfat renaming support

leigh123linux:
Fixes:  https://github.com/linuxmint/nemo/issues/2708


eyalroz:
So what exactly changes with this PR merged?

leigh123linux:
@eyalroz  It adds support for kernel exfat.
Microsoft files systems can't handle some characters, see

https://github.com/linuxmint/nemo/blob/def5cf21c744e7926de8828679ed101f3940d885/libnemo-private/nemo-file-operations.c#L3393

eyalroz:
@eyalroz :  I know, I filed #2708 ; I mean, in what sense is support added? What happens now when you copy to/from exfat and some characters are unsupported?

mtwebster:
Any invalid characters are replaced with an underscore `_`.

eyalroz:
@mtwebster : Ah, ok, now I get it. Would have spared you the question if you would have been this verbose in the original PR description :-P

Issue #2970, opened 2022-04-24
Sort folders before files preference doesn't affect dconf /org/gtk/settings/file-chooser/sort-directories-first


JamiKettunen:
```
 * Nemo version - 5.2.4
 * Issue is with - Windowed nemo
 * Distribution - Void Linux
 * Graphics hardware *and* driver used - AMD RX 580 with amdgpu kernel driver
 * 64 bit
 ```

**Issue**
Changing the `Sort folders before files` preference doesn't change the GTK `/org/gtk/settings/file-chooser/sort-directories-first` dconf entry value for e.g. save/open file dialogs.

**Steps to reproduce**
1. Check `dconf read /org/gtk/settings/file-chooser/sort-directories-first`
2. Open nemo
3. Change `Edit` > `Preferences` > `Sort folders before files`
4. Check `dconf read /org/gtk/settings/file-chooser/sort-directories-first` and observe it hasn't changed
5. Trying to open/save files from other applications don't respect the value of the `Sort folders before files` preference chosen in nemo

**Expected behaviour**
Changing `Sort folders before files` preference also affects the system-wide GTK open/save dialogs.

[Nautilus](https://wiki.gnome.org/action/show/Apps/Files) already links it's "sort folders before files" preference to this dconf/gsettings value and I think it would simply make sense to do the same here.

**Other information**
Fwiw the state of the GTK `sort-directories-first` boolean can also be get/set via `gsettings`, e.g. `gsettings get org.gtk.Settings.FileChooser sort-directories-first`

**Comment by leigh123linux** [comment]
Nemo uses org.nemo.preferences sort-directories-first

```
gsettings get org.nemo.preferences sort-directories-first
```

JamiKettunen:
I know it uses that but my issue is it doesn't affect the GTK file picker and makes nemo feel unpolished as a result.

So will the nemo preference never be linked to `org.gtk.Settings.FileChooser sort-directories-first` then?

JamiKettunen:
@leigh123linux Could I please get a second opinion on this issue from another developer? I think it's a bit rude to close this when what you said didn't even address my issue in any way.

leigh123linux:
> @leigh123linux Could I please get a second opinion on this issue from another developer? I think it's a bit rude to close this when what you said didn't even address my issue in any way.

Perhaps you should make your reports clearer, you never mentioned you expected the setting to be linked.
As for calling me rude, FUCK YOU!

PR #2971, opened 2022-04-25
Sync sort-directories-first with gtk file chooser

leigh123linux:
Fixes: #2970


leigh123linux:
Show hidden files could also be added later, see

https://github.com/GNOME/nautilus/commit/85524e01c37ef6cfa617190fe77b1a8381b31080

PR #2986, opened 2022-05-16
Update nemo-autostart.desktop.in

dennisr40:
I use Nemo with GNOME not Cinnamon. Why should hide the icon?


SparkyBluefang:
Yeah, if you're mixing and matching components from different desktop environments (i.e. GNOME, Cinnamon, Mate, KDE, etc) then this is something you should expect you'll need to deal with. You should either locally edit your `.desktop` files or create your own (so they don't get overridden by updates).

I'd say this is functioning as intended and desired.

PR #2995, opened 2022-05-29
adding https to web url

Yofenry:
Affected Line = 2
Changes = Original:
<its:rules xmlns:its="http://www.w3.org/2005/11/its"
New:
<its:rules xmlns:its="https://www.w3.org/2005/11/its"


mtwebster:
Hi this is standard boilereplate for that type of file, so it should be left alone.

Issue #3012, opened 2022-07-03
Nemo actions does not escape backslashes in paths


veractor:
```
 * Nemo version (nemo --version): nemo 5.2.4
 * Is issue with desktop or windowed nemo? Both
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...): Fedora 36
 * 32 or 64 bit: 64 bit
 ```

**Issue**

Nemo actions does not escape backslashes in paths at all, making Nemo actions target the wrong file if the selection filename contains an escape sequence.

**Steps to reproduce**

1. Save the following as `~/.local/share/nemo/actions/print_args.nemo_action`:
```
[Nemo Action]
Active=true
Name=Print arguments
Comment=Print args
Exec=<print_args.py %F>
Selection=notnone
Extensions=any;
```
2. Save the following as `~/.local/share/nemo/actions/print_args.py` and mark it as executable:
```python
#!/usr/bin/python3

import sys

print("Arguments:")
for i in range(len(sys.argv)):
    print(f"{i}: " + sys.argv[i])
```
3. Run `nemo --quit; NEMO_ACTION_VERBOSE=1 nemo` in a terminal.
4. Create a file with filename `` \`.txt ``, and a file with filename `` `.txt ``.
5. Right click the first file, and select "Print arguments". Then same with the second file.
6. The terminal outputs:
```
Arguments:
0: /home/veractor/.local/share/nemo/actions/print_args.py
1: /home/veractor/`.txt
Arguments:
0: /home/veractor/.local/share/nemo/actions/print_args.py
1: /home/veractor/`.txt
```

**Expected behaviour**

The terminal should output:
```
Arguments:
0: /home/veractor/.local/share/nemo/actions/print_args.py
1: /home/veractor/\`.txt
Arguments:
0: /home/veractor/.local/share/nemo/actions/print_args.py
1: /home/veractor/`.txt
```

**Other information**

This would still happen when `Quote` is set to `double`. Setting it to `single` fixes it, but if the filename contains an odd number of single quotes, then the command would fail to trigger, and if the filename contains an even number of single quotes, then the single quotes are removed (See #798). Exact same issue occurs when putting single or double quotes around `%F` in the `Exec` line instead. Setting `Quote=backtick` has no effect other than adding 2 extra backticks.

PR #3018, opened 2022-07-08
Actions: Escape tabs, newlines and backslashes

### LINKED ISSUE #3011: Nemo actions does not escape tabs and newlines in paths
```
 * Nemo version (nemo --version): nemo 5.2.4
 * Is issue with desktop or windowed nemo? Both
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...): Fedora 36
 * 32 or 64 bit: 64 bit
 ```

**Issue**

Nemo actions do not escape tabs and newlines in paths at all, making Nemo actions on files with filenames that contain those characters impossible while also handling quotes in filenames.

**Steps to reproduce**

1. Save the following as `~/.local/share/nemo/actions/print_args.nemo_action`:
```
[Nemo Action]
Active=true
Name=Print arguments
Exec=<print_args.py %F>
Selection=notnone
Extensions=any;
EscapeSpaces=true
```
2. Save the following as `~/.local/share/nemo/actions/print_args.py` and mark it as executable:
```python
#!/usr/bin/python3

import sys

print("Arguments:")
for i in range(len(sys.argv)):
    print(f"{i}: " + sys.argv[i])
```
3. Run `nemo --quit; NEMO_ACTION_VERBOSE=1 nemo` in a terminal.
4. Create a file with filename `Newline:<newline>Tab:<tab>.txt`. (The `<newline>` and `<tab>` is a literal newline and literal tab)
5. Right click the file, and select "Print arguments".
6. The terminal outputs:
```
Arguments:
0: /home/veractor/.local/share/nemo/actions/print_args.py
1: /home/veractor/Newline:
2: Tab:
3: .txt
```

**Expected behaviour**

The terminal should output:
```
Arguments:
0: /home/veractor/.local/share/nemo/actions/print_args.py
1: /home/veractor/Newline:
Tab:	.txt
```

**Other information**

Setting `Quote=double` does not work since filenames with an odd number of double quotes would then fail to trigger the command in `Exec`, and filenames with an even number of double quotes would have the quotes removed. Same with `Quote=single` for single quotes (This is what #798 is). Exact same issue occurs when putting single or double quotes around `%F` in the `Exec` line instead. Setting `Quote=backtick` has no effect other than adding 2 extra backticks.

veractor:
Fixes #3011 and #3012. Tabs, newlines and backslashes is set to have addition processing if the path is not quoted at all.

### DIFF
--- a/eel/eel-string.c
+++ b/eel/eel-string.c
@@ -106,7 +106,7 @@ eel_str_escape_spaces (const char *string)
 char *
 eel_str_escape_non_space_special_characters (const char *string)
 {
-    int quotes_and_tabs;
+    int quotes_tabs_and_backslashes;
     int newlines;
     const char *p;
     char *q;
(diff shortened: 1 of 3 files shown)


mtwebster:
Thanks - btw,

I noticed you had NEMO_ACTION_VERBOSE for debugging - that's been removed, and actions have been made a debug topic:
```
NEMO_DEBUG=Actions nemo --debug
```
You can run with `NEMO_DEBUG=help` to see other topics.

veractor:
Ah, I think I originally used that but switched to the older way that I copied from some old issue because it had cleaner output. Didn't know that was removed, thanks for the tip.

veractor:
Also, any clue as to why the EscapeSpaces option exist? I feel like the paths for actions should be escaped by default when no quoting is applied, so that no additional options should be needed to handle pathnames with spaces.

veractor:
Ah ok I understand. I hope my changes didn't affect any existing behaviour then.

PR #3021, opened 2022-07-13
Free grid alignment of the thumbnails

cestlotre:
Would it be possible to implement in Nemo a free grid alignment of the thumbnails?
I mean that moving a thumbnail into the free space of the folder is possible as in the attached illustration (this is "Folder Options X" software in Windows 8.1).
This is a free alignment of the thumbnails but with the constraint of alignment on the grid.

Capture


mtwebster:
Right-click on a blank area, Arrange Items->Manually?

PR #3025, opened 2022-07-16
Add Sevelopment Setup Commands

dparthka:


mtwebster:
Hi, I wouldn't mind having some build instructions, but it needs to be a bit better than this - check out https://github.com/linuxmint/xviewer or https://github.com/linuxmint/xreader/blob/master/INSTALL.md

PR #3041, opened 2022-07-31
mime-application-chooser: Quote custom executable file paths

veractor:
Currently, when selecting a custom executable file in the application chooser, the spaces in its path is escaped, but that's it. This means that if an executable file were to be chosen, with a path of say, `/home/veractor/Katie's scripts/transform.sh`, the text box that would end up with the path would show `/home/veractor/Katie's\ scripts/transform.sh`, and the file would be reported as being "Not an executable file", until the `'` is manually escaped with a backslash.

Always quote the chosen path with GLib's `g_shell_quote` function instead of just escaping the spaces so that navigating to the executable file would work directly, instead of having to escaping the path additionally.


mtwebster:
Hi can you get rid of the translation changes so it's just the fix?

I wouldn't mind changing the wording in the future, but it will break current translations and that's not something I want to do at this point in our 5.4 release.

Thanks

veractor:
Done!

PR #3086, opened 2022-09-30
File extension column 

VitanovG:
This pull request integrates the File extension column into nemo. You can use the column in list view, and you can sort the files in any views according to file extensions. This is useful in case someone works with Latex and has 10+ .tex files, 10+ .sty files and 10+ .aux files in a directory, and would like to sort them such that:
1.aux
2.aux
3.aux
a.sty
b.sty
1.tex
2.tex
3.tex

will be the sort behavior. This is not possible with the current "Type", "Detailed type" and "MIME type" methods. This feature was already requested in the issues, and dolphin recently added the same "Extension" column.


Drugwash2:
Guess this option has never been considered before by the official developers because "Linux is not Windows" - which is the default motivation for everything that looks and/or works in an abnormal way as seen by former Windows users (or people that have some logic in their thinking).

What needs done next is adding an option for displaying file size **in bytes only**, with thousand separators, because both current options are either uninformative or space wasting.

Then it desperately needs a "common-sense" sorting order, which is based on the old ANSI character set and has always been the default in Windows. Some call it natural sorting. Back when I was still struggling to use Nemo it was making me crawl up the walls when subfolders named as `_test` or `__work` (note the underscores!) were always found aligned towards the bottom of the list with T-named or W-named subfolders, respectively, instead of being at the very top of the current folder/partition where I had always expected them to be for the past 20+ years.
Apparently others do have enough common-sense to provide choices for this, for example Double Commander:
Screenshot from 2022-09-30 16-09-06

libnemo-private/nemo-file.c
```diff
@@ -4036,6 +4065,28 @@ nemo_file_peek_display_name_collation_key (NemoFile *file)
...
+	char *str, *token, *result;
+	str = g_strdup (eel_ref_str_peek(file->details->name));//strdup()
+	
+	while ((token = strsep(&str, "."))) {result = g_strdup(token);}
+	if (result == NULL) {result = "";}
+	g_free(str);
+	g_free(token);
+	
+	return result;
+}
```
mtwebster, line 4088: There are a few problems here:

Remember you're not returning a new string (`const char*`), since you're 'peeking' at the value:
- link broken string should just be a static string (no g_strdup).
- The remaining code either returns a new string, or a static empty one, and leaks memory (you never free `token` while running the `strsep` loop, and if there are multiple separators the intermediate ones leak).

I would suggest using `g_strrstr` as it works from the end of the string and returns a pointer or `NULL`, and no new strings need to be created:
```
    const char *str, *ptr;
    str = eel_ref_str_peek (file->details->name);

    if ((ptr = g_strrstr (str, "."))) {
        return ptr + 1;
    }

    return "";
```
We usually try to stick with glib functions, as they're generally more convenient (and often safer).

libnemo-private/nemo-file.c
```diff
@@ -4077,6 +4128,12 @@ nemo_file_get_display_name (NemoFile *file)
 	return g_strdup (nemo_file_peek_display_name (file));
 }
 
+char *
+nemo_file_get_extension_name (NemoFile *file)
+{
+	return nemo_file_peek_extension_name (file);
+}
```
mtwebster, line 4135: Here you're returning a new string (`char *` return), so you would need to g_strdup the peeked extension name.

mtwebster:
https://github.com/linuxmint/nemo/commit/a89c95c108cdb1b7d2e62b95f37891c51ae7e840

Issue #2834, opened 2021-09-19
Port to tracker-3


pacho2:
```
 * Nemo version (nemo --version) 5.0.3
 * Is issue with desktop or windowed nemo? All
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) Gentoo
 * Graphics hardware *and* driver used
 * 32 or 64 bit 
 ```

**Issue**
Even with 5.0.3 version, it seems that nemo is not ported to tracker-3, being one of the last packages still requiring old tracker-2
https://gitlab.gnome.org/GNOME/Initiatives/-/issues/17

It would be nice to try to port it if possible

Thanks a lot

mtwebster:
Most people don't build with tracker support so it's not a priority (other than making sure we don't break it). It looks like it should be fairly easy to port, I'll take a look at it.

Dr-Bone:
Hi
I'm using Nemo and Cinnamon on Arch Linux. 
Tracker has gone from the official repos, tracker3 it is it seems. Is it still planned to integrate tracker3?
Or is it possible already and am I just missing a config?
thanks!

PR #3093, opened 2022-10-07
[Next] Port to tracker-sparql 3 (#2834)

ItzSwirlz:
Mint 21 is now based on Ubuntu 22.04, and it contains sparql 3

We basically don't need it's only usage of the removed function. I did what GNOME did: https://github.com/GNOME/nautilus/blob/master/src/nautilus-search-engine-tracker.c

### DIFF
--- a/libnemo-private/nemo-search-engine-tracker.c
+++ b/libnemo-private/nemo-search-engine-tracker.c
@@ -362,21 +362,5 @@ nemo_search_engine_tracker_init (NemoSearchEngineTracker *engine)
 NemoSearchEngine *
 nemo_search_engine_tracker_new (void)
 {
-	NemoSearchEngineTracker *engine;
-	TrackerSparqlConnection *connection;
-	GError *error = NULL;
-
-	connection = tracker_sparql_connection_get (NULL, &error);
-
-	if (error) {
-		g_warning ("Could not establish a connection to Tracker: %s", error->message);
-		g_error_free (error);
-		return NULL;
-	}
-
-	engine = g_object_new (NEMO_TYPE_SEARCH_ENGINE_TRACKER, NULL);
-	engine->details->connection = connection;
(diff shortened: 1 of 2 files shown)


Fantu:
Thanks for your working for porting to newer softwares/versions
I think you can remove also tracker-sparql-1.0 support (is very old, removed from debian 10, nautilus removed for example 4 years ago: https://github.com/GNOME/nautilus/commit/890760bbce7585943ac3eed9375eddd19ef4b932)

And are you sure no other changes are needed? For example the port to tracker3 in nautilus: https://github.com/GNOME/nautilus/commit/7712c2c7e10b3c28b2cc2ec8dfeeb8a89bd57945
And with changes not compatible with tracker 2 will require also remove its support

Anyway I didn't a deep search in nemo and part related to tracker, these days I am also sick

ItzSwirlz:
I'll have to do a look-through everything - i'll do some investigation and see what the best thing to do is

libnemo-private/nemo-search-engine-tracker.c
```diff
@@ -362,21 +362,5 @@ nemo_search_engine_tracker_init (NemoSearchEngineTracker *engine)
...
-		g_error_free (error);
-		return NULL;
-	}
-
-	engine = g_object_new (NEMO_TYPE_SEARCH_ENGINE_TRACKER, NULL);
-	engine->details->connection = connection;
-	engine->details->cancellable = g_cancellable_new ();
-
-	return NEMO_SEARCH_ENGINE (engine);
+	return (NEMO_TYPE_SEARCH_ENGINE_TRACKER, NULL);
```
okaestne, line 365: I think there is a `g_object_new` missing

mtwebster:
https://github.com/linuxmint/nemo/commit/b03e84191b4256d12b14bbbb5d7d6d9b78dc164f

PR #3131, opened 2022-12-04
Fix memleaks

okaestne:


src/nemo-places-sidebar.c
```diff
@@ -1168,6 +1168,7 @@ update_places (NemoPlacesSidebar *sidebar)
 
         mount = g_volume_get_mount (volume);
         if (mount != NULL) {
+            g_autofree gchar *parse_name;
```
mtwebster, line 1171: Can you initialize this to NULL, or move this declaration down and initialize it with g_file_get_parse_name).

okaestne: thanks for pointing this out, fixed and rebased

PR #3188, opened 2023-02-19
Fix List View icon size vs. zoom level

kainzilla:
Currently List View displays icons one level of zoom smaller than the actual zoom level in use - this should fix that issue.


mtwebster:
I'm not sure I like how this affects the icon/text relationships. It's been too long to remember, but I have a feeling the current sizes are this way on purpose - 

see:
36d2b30d28a69c6151abbbe0112acb86b36992fa
followed by
1dbd08c4145c0e7888e89a2f32f3ffc59b3b1426

kainzilla:
Understood, and that looks right - I'll go ahead and close this request!

PR #3220, opened 2023-04-15
Use local URIs for drag&drop, copy/paste operations

Guldoman:
This is done to accommodate applications that don't support parsing special URIs.

This converts paths like `sftp://...` into their local equivalent, like `file:///run/user/1000/gvfs/sftp:...`.

### DIFF
--- a/libnemo-private/nemo-clipboard-monitor.c
+++ b/libnemo-private/nemo-clipboard-monitor.c
@@ -233,7 +233,7 @@ convert_file_list_to_string (NemoClipboardInfo *info,
 	}
 
         for (i = 0, l = info->files; l != NULL; l = l->next, i++) {
-		uri = nemo_file_get_uri (l->data);
+		uri = nemo_file_get_local_uri (l->data);
 
 		if (format_for_text) {
 			f = g_file_new_for_uri (uri);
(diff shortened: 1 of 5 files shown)


mtwebster:
This doesn't seem to be changing any behavior for me - how are you testing?

If I:
- Drag a file from nemo over a terminal, I get a local path (with or without this patch)
- If I copy a file, then paste into a terminal, I still get the uri, *even though nemo_file_get_local_uri() appears to be working correctly*.  Again, this is the same with or without this patch.

Guldoman:
For dnd I'm using an SDL test (`testdropfile`) and an SDL-based text editor (https://github.com/lite-xl/lite-xl).

For copy-paste you can test with `xclip -t text/uri-list -o -selection clipboard`.

Just to be clear, you should always get a uri, but it should be one in the format of `file://` even for remote mounts.

For *textual* copy-paste, I left the "original" uri as the local one can be quite long in comparison.

Guldoman:
Just to clarify, the behavior for dnd that this PR implements is the same for Thunar (https://github.com/xfce-mirror/thunar/commit/5f1c0e8974a7680fa45b4cfbd1ba398d303eeb24) and Nautilus (https://gitlab.gnome.org/GNOME/nautilus/-/commit/1d1488cbc96ee9f7dd69b708e093d274a5374878 and https://gitlab.gnome.org/GNOME/nautilus/-/commit/9adf6760c6e2546378630d787e82c3d514e70c61).

With the latest change, URIs like `recent://` are supported too.

PR #3225, opened 2023-04-26
properties-window: split dir elem count and size

### LINKED ISSUE #2786: [Feature request] Make the file properties screen also show the size in bytes regardless of the prefix setting
```
 * Nemo version (nemo --version) - 5.0.3
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) - Mint 20.2 Uma
 ```

**Issue**

I often use the "Binary" prefix because it takes up little screen space and doesn't cause confusion when looking quickly at the size column (in the list view). But it is extremely inconvenient to have to change the prefix to "Binary (long format) in order to see the size in bytes of a file/folder. 

**Expected behaviour**

I would like the file properties screen to always show the size of the file/folder in bytes next to  the size in the prefix chosen by the user.

Thank you.

### PR DESCRIPTION
**Pull request by okaestne**
In the properties window of multiple files and directories, split the number of elements and their total size into separate rows. Also always display the (total) file size in the long format (i.e. incl. the number of bytes).

This is quite hacky and there are many special cases, so please let me know if you have any suggestions or improvements.

Fixes #2786

### DIFF
--- a/src/nemo-properties-window.c
+++ b/src/nemo-properties-window.c
@@ -108,6 +108,7 @@ struct NemoPropertiesWindowDetails {
 
 	GtkLabel *directory_contents_title_field;
 	GtkLabel *directory_contents_value_field;
+	GtkLabel *size_value_field;
 	guint update_directory_contents_timeout_id;
 	guint update_files_timeout_id;
 
@@ -194,6 +195,9 @@ static const GtkTargetEntry target_table[] = {
  */
 #define CHOWN_CHGRP_TIMEOUT			300 /* milliseconds */
 
+static gboolean is_computer_directory             (NemoFile *file);
+static gboolean is_network_directory              (NemoFile *file);
+static gboolean is_burn_directory                 (NemoFile *file);
 static void directory_contents_value_field_update (NemoPropertiesWindow *window);
 static void file_changed_callback                 (NemoFile       *file,
 						   gpointer            user_data);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/nemo-properties-window.c [hunk 1]
```diff
@@ -2148,6 +2164,21 @@ directory_contents_value_field_update (NemoPropertiesWindow *window)
 			    text);
 	g_free (text);
 
+	/* Update the file size field with the selection or directory's total file size. */
+	if (is_computer_directory (file) ||
+	    is_network_directory (file)) {
```
mtwebster, line 2169: You're using `file` here which isn't initialized except inside of a loop at the beginning of `directory_contents_value_field_update()` - so it seems like you'll end up always evaluating the last file of the selection.

You could maybe check this for each file inside the loop, and if any are the pseudo file-type you can act on it here. Or maybe you only need to check selection's parent directory?

mtwebster:
One other minor thing I noticed:
image

The size row should be hidden for empty folders (this is its current behavior I think).

src/nemo-properties-window.c
okaestne: Interesting.. In case of a multi-selection, the var `total_size` is calculated in the first `for` loop, which is then used to display the file size. In case a single directory is selected, the for loop initializes `file` with this exact directory.

So as long as you don't have the special `computer:///` or `network:///` "directories" in a multi-selection, this should even work as intended :smile:

okaestne: @mtwebster I changed the code a bit to make it more clear

okaestne:
> The size row should be hidden for empty folders (this is its current behavior I think).

Thing is, the `GtkLabel`s "Size:" and the value placeholder are added to the `GtkGrid` before the contents and size are determined. So I'd need to store both labels and remove them from the grid, if the calculation finishes with result "one empty folder". Also I think "Size: 0 bytes" makes it more clear, that the folder is empty.

Issue #2979, opened 2022-05-02
Nemo thumbnail / preview generation fails on samba share


pdev77:
```
 * Cinnamon version = Cinnamon 5.2.7
 * Distribution - Linux Mint 20.3 Una
 * Graphics hardware *and* driver used
 * 64 bit
(nemo:3703): Gtk-WARNING **: 22:12:38.206: Failed to register client: GDBus.Error:org.gnome.SessionManager.AlreadyRegistered: Unable to register client

(nemo:3703): CinnamonDesktop-WARNING **: 22:13:06.253: Unable to create loader for mime type video/mp4: Unrecognised image file format

(nemo:3703): CinnamonDesktop-WARNING **: 22:13:06.254: Error creating thumbnail for smb://192.168.1.11/photos2022_rw/ToProcess/VID-20210224-WA0004-(2MB).mp4: Unrecognised image file format

 ```
**Issue**
Unable to generate thumbnails for ANY mp4 files on samba mount
Those same files can have thumbnails generated on local drive or when browsing the files via the /var/run/user/1000/gvfs/smb-share:server=192.168.1.11.. mount

**Steps to reproduce**
Launch nemo
Browse to "smb://192.168.1.11/photos2021_rw/"
mp4 files do not generate thumbs (setting for preview in nemo is "YES" and for files < 1GB
Video file in question is 6MB

Browse to /var/run/user/1000/gvfs/smb-share:server=192.168.1.11,share=photos2021_rw
Thumbs are generated

**Expected behaviour**
Thumbs are generated

**Other information**
Same behaviour replicated from 20.3 live USB image

**Comment by pdev77** [comment]
* VGA compatible controller: Intel Corporation HD Graphics 530 (rev 06) / i915 . Kernel 5.

**Comment by vimes666** [comment]
I just handled a complaint about missing thumbnails when connecting to a phone on the linux mint forum.
I noticed the 'show-image-thumbnails' setting in the nemo preferences has been removed.
The default was 'local files only' which corresponds with the observed behaviour.
I supplied a work-around using the dconf-editor in the forum topic. Here is the topic: [](https://forums.linuxmint.com/viewtopic.php?f=49&t=388423)

**Comment by PS1Online** [comment]
The problem persists on Nemo 5.6.3 (latest version)

**Comment by trakout** [comment]
Also have this issue on `5.6.4`, but the following workaround works well and results in video thumbnails loading with Nemo:

Create a mount point:
```
[...]

PR #3265, opened 2023-07-09
Fix issue when generating video thumbnails from network shares

endoedgar:
I've found that Nemo wasn´t generating video thumbnails whenever I was in a network share (I was using the default ffmpegthumbnailer). I was finding it weird because image thumbnails were fine.

Digging into the code I've found that we are passing the smb:// file uri to the gnome_desktop_thumbnail_factory_generate_thumbnail cinnamon-desktop function, and that in turn is calling g_filename_to_uri whenever it finds a %i in a thumbnailer. 

That was exactly the case with ffmpegthumbnailer. Image thumbnailers normally don't use the %i format, but %u instead.

Since g_filename_to_uri doesn´t support anything else than local file:// uri, I've added this code to take care of that and fix the thumbnail generation.

### DIFF
--- a/libnemo-private/nemo-thumbnails.c
+++ b/libnemo-private/nemo-thumbnails.c
@@ -318,6 +318,8 @@ thumbnail_thread (gpointer data,
     NemoThumbnailInfo *info = (NemoThumbnailInfo *) data;
     GdkPixbuf *pixbuf;
     time_t current_time;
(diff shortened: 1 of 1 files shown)


endoedgar:
This will probably fix this reported issue https://github.com/linuxmint/nemo/issues/2979

PR #3269, opened 2023-07-19
Use `activation_uri` as a second choice in `nemo_file_get_local_uri`

Guldoman:
Using `activation_uri` can be problematic with links.

I can't test the "desktop" behavior, but maybe with this change the `NEMO_IS_DESKTOP_ICON_FILE` isn't needed anymore.

Ref: https://github.com/linuxmint/nemo/commit/84b6062319f4029e0e99b9eb28c77e22c2bf9bb8#r122103578


mtwebster:
Alright we'll try this, I haven't been able to break it yet

PR #3281, opened 2023-07-26
eel-gtk-extensions.c: (Wayland) Fix menu popup positioning, menu closing issues.

### LINKED ISSUE #3218: BUG: Using Wayland - Context Menu (via Right-Click) Doesnt Close when Clicking Away on Left Side
This problem only happens using Wayland (in Ubuntu). So, this is not an X11 problem and therefore doesn't affect Mint (and, by extension, Cinnamon) users.

This has been a bug for years in Wayland.  Please fix!

position4

**How to Recreate the Problem:**

1. Using GNOME (with Wayland as the windowing protocol) open a file window in list view.
2. Right-click directly on an existing file to bring up its context menu. You **_MUST_** directly highlight the file for this problem to occur (in the video it's shown in orange).
3. Click on the left-side, outside of this context menu.  It won't close!

You would expect this context menu to close. It won't.   There appears to be an invisible window on the left-side of this context menu with  the same approximate dimensions of the visible context menu.  It is there and does not register click events.  It's very jarring when trying to click-close.  You always think Nemo crashed hard.

The only way to close this menu (besides selecting an item in this menu) is to either click above, below, right-side, or move a distance away on the left-side.

### LINKED ISSUE #3280: Context menus show up in the middle of screen
### Distribution

Arch Linux 6.4.4.arch1-1

### Package version

5.8.4-1

### Frequency

Always

### Bug description

Context menus show up in the middle of screen when using keyboard controls and the mouse wasn't on top of nemo's window.

### Steps to reproduce

1. Press context menu key on the keyboard to show context menu
2.1. If there was a mouse pointer over the window, the context menu will show up under **mouse pointer** (Not the cursor!)
2.2. If there was no mouse in view, a small floating window with context menu shows up in the center of **screen** (not window's center)

### Expected behavior

Context menus should display under cursor when not using mouse. Similarly to `xdg-desktop-portal-gtk`, no matter if the mouse is in view or not, specifically when using context menu key on the keyboard.

### Additional information

Tested on Hyprland and River.
Hyprland has a rare chance where there won't be a floating window at all (when reproducing steps 1. and 2.2.) and currently selected option will only be displayed in status bar.

River doesn't share this uncommon behavior.

mtwebster:
When opening the menu without the mouse (shift-F10, keyboard context  menu key) gtk_menu_popup complains about not having a parent, and pops up in the wrong location.

Using gtk_menu_popup_at_rect() with a fabricated rectangle based on the pointer location works instead.

A pleasant side effect of this is that it works around another bug where clicking off the left side of a popup fails to close the menu or change focus at all. The menu now closes fine, regardless of where the user clicks.

In X11 sessions gtk_menu_popup_at_pointer() is still used.

Fixes #3218

ref: https://bugs.kde.org/show_bug.cgi?id=453532

### DIFF
--- a/eel/eel-gtk-extensions.c
+++ b/eel/eel-gtk-extensions.c
@@ -37,6 +37,7 @@
 #include <X11/Xatom.h>
 #include <gdk/gdk.h>
 #include <gdk/gdkprivate.h>
+#include <gdk/gdkwayland.h>
 #include <gtk/gtk.h>
 #include <glib/gi18n-lib.h>
 #include <math.h>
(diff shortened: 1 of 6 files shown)


Lassebq:
Fixes #3218 but not #3280. Context menus still behave weirdly and are either opened at that last mouse pointer pos or in the top left corner. I'd recommend checking out xdg-desktop-portal-gtk, it shows a pop up like this.
image

Lassebq:
But other than that I think it's good to be merged

Issue #3152, opened 2023-01-02
[moving files] stuck on conflict "duplicate" option


ToxicMushroom:
```
 * Nemo version: 5.6.1
 * Is issue with desktop or windowed nemo?: Not sure what this means, nemo is being windowed yes.
 * Distribution: Arch
 * Graphics hardware *and* driver used: AMD RX6700XT | mesa
 * 64 bit
 ```

**Issue**
Cannot move file named "example" into a folder which has files  "example" and "example (copy)".

**Steps to reproduce**
`touch file1`
`mkdir folder1`
`cp file1 folder1`
now use nemo to copy file1 into folder1, choosing duplicate as conflict solution (this will succeed)
now use nemo to move file1 into folder1, choosing duplicate as conflict solution (this will hang forever)

**Expected behaviour**
nemo find the next available filename and moves the file.

**Other information**
I am using an ssd, comparable file moves are instantanious when no conflicting names are present.
here are some illustrative images
![image](https://user-images.githubusercontent.com/32853531/210284123-150f7749-88b7-475a-9a1a-3f861015e398.png)
![image](https://user-images.githubusercontent.com/32853531/210284139-ee702160-69ee-4fd7-a56c-ffb47df62333.png)

MegaV0lt:
Same issue here.

EXtremeExploit:
just for a bit of extra info, this only happens on moving, not on copy, so my guess is that some extra checking for the next available filename is not done on move but that its present on copy

PR #3299, opened 2023-08-16
Fix trying the same filename when moving a file with duplicate name

EXtremeExploit:
fix https://github.com/linuxmint/nemo/issues/3152
So basically the count it was trying for was hardcoded to one, meaning it will always try the "filename (copy)" filename instead of incrementing, never completing and waiting forever, PR makes it so that it will actually track how many times it failed and try the next iteration of names, tried to make it just like the other operations


carlosmintfan:
@mtwebster Could you backport this to older mint versions, for example, 20.3? I also encounter the issues there.

PR #3307, opened 2023-09-04
nemo-view.c: Show detailed actions for some selection types

rcalixte:
Discussion: linuxmint/discussions#114

### DIFF
--- a/src/nemo-view.c
+++ b/src/nemo-view.c
@@ -305,6 +305,8 @@ struct NemoViewDetails
     GVolumeMonitor *volume_monitor;
 
     GTimer *load_timer;
+
+    char *detail_string;
 };
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


src/nemo-view.c
```diff
@@ -3133,6 +3135,7 @@ nemo_view_display_selection_info (NemoView *view)
 			}
 
 		}
+		view->details->detail_string = g_strdup (folder_item_count_str);
```
mtwebster, line 3138: Need to `g_free` any previous contents of `view->details->detail_string`.

It also needs to be freed in `nemo_view_finalize` (g_clear_pointer)

src/nemo-view.c
```diff
@@ -3171,11 +3174,14 @@ nemo_view_display_selection_info (NemoView *view)
 							  items_string,
 							  size_string);
 
+
+			view->details->detail_string = g_strdup_printf (_(" (%s)"), size_string);
```
mtwebster, line 3178: same

src/nemo-view.c
```diff
@@ -6820,17 +6827,17 @@ copy_or_cut_files (NemoView *view,
 		if (cut) {
 			status_string = g_strdup_printf (ngettext("The %'d selected item will be moved "
 								  "if you select the Paste command",
-								  "The %'d selected items will be moved "
+								  "The %'d selected items%s will be moved "
```
mtwebster, line 6830: this will require re-translation, maybe some explanation for the potfile on the line above for here and the one below. (A comment just above the gettext call will be automatically included for translators).

PR #3334, opened 2023-12-06
Fix order of actions in context menu

zsugabubus:
Opening `nemo 6.0.0` (but issue first seen about half year ago so it is not version specific), in the first tab of the first window, actions in context menu show up in reversed order. Order seems to be correct in further tabs and windows or after changing view mode (e.g. list -> icons).

### DIFF
--- a/libnemo-private/nemo-action-manager.c
+++ b/libnemo-private/nemo-action-manager.c
@@ -145,37 +145,13 @@ sort_file_list_cb (gconstpointer a, gconstpointer b)
     NemoFile *file_b;
     const char *s0;
     const char *s1;
-    gint r;
 
     file_a = (NemoFile *) a;
     file_b = (NemoFile *) b;
     s0 = nemo_file_peek_name (file_a);
     s1 = nemo_file_peek_name (file_b);
-    //
-    // Order alphabetically
-    // Compare ASCII values of file names char by char
-    //    < 0:  a < b
-    //   == 0:  a == b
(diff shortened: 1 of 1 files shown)


zsugabubus:
Sorry if I was not totally clear, my issue was not about global ordering but order of actions from a single directory.

```
cd ~/.local/share/nemo/actions
/path/to/helper.zsh 3
```

<details>

<summary>
helper.zsh
</summary>

```zsh
#!/bin/zsh -xeu
n=$1

# Will remove your existing actions, be careful!
# rm -f *.nemo_action(N); # Uncomment me.

for i in {001..$n}
do
  printf > $i.nemo_action "
[Nemo Action]
Name=x-%s
Exec=echo
Selection=any
Extensions=any
" "$i"
done
```

</details>

I'm on Arch Linux, using i3 (no desktop environment). The order I see in context menu is 3 2 1. In further tabs it correctly shows 1 2 3.

I guess it is likely related to timings of events and the number of directories. On my machine I can reproduce it with 100%. I can provide a similar output like yours if it would be helpful.

A final sorting would fix a superset of this issue so maybe I should give a try and implement that one too?

zsugabubus:
What's the state of this PR? Am I expected to do something?

mtwebster:
Sorry, nothing really wrong, it's just a bit incomplete like I mentioned above.

This will be less of an issue when https://github.com/linuxmint/nemo/pull/3353 is merged also.

PR #3340, opened 2023-12-28
nemo-desktop: use app_id as .desktop file name

sertonix:
This is suggested in the [xdg-shell spec](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/463d5cfaf37c8276826cdda48df336b9d2511ed0/stable/xdg-shell/xdg-shell.xml#L651) and makes finding the `.desktop` file from an application easier.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -18,7 +18,7 @@ desktopFiles = [
 
   configure_file(
     input : 'nemo.desktop.in',
-    output: 'nemo.desktop',
+    output: 'org.Nemo.desktop',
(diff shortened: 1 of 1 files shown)


clefebvre:
This was reverted. We don't have a DNS for this project, let alone a reverse DNS. org.Nemo is completely random and harder to find than the appname.

sertonix:
I wanted to make sure that the desktop file name matches the app_id but I see that you changed the app_id accordingly so I am fine with that revert.

PR #3349, opened 2024-01-18
Support drawing the desktop on wayland using gtk-layer-shell.

lcolitti:
This PR allows nemo-desktop to display desktop icons on wayland compositors that support the [layer-shell](https://wayland.app/protocols/wlr-layer-shell-unstable-v1) protocol.  It adds an optional dependency on gtk-layer-shell, which supports wlroots-based compositors such as sway, wayfire etc. If the dependency is found, then nemo will run natively on wayland.

I have been using it on [wayfire](https://wayfire.org/) and it works well.

The change is very simple and should be easy to maintain even if Wayland is not yet officially supported.

### DIFF
--- a/config.h.meson.in
+++ b/config.h.meson.in
@@ -19,6 +19,8 @@
 #mesondefine HAVE_SELINUX
 // Define to enable pango-1.44 fixes
 #mesondefine HAVE_PANGO_144
+// Define to use gtk-layer-shell
+#mesondefine HAVE_GTK_LAYER_SHELL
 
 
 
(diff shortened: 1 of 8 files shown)


Fantu:
For use it in mint packages should be added in debian/control (build-dep) and in debian/rules (configure parameter)

lcolitti:
> For use it in mint packages should be added in debian/control (build-dep) and in debian/rules (configure parameter)

Done, I think. I built a debian package and installed it and desktop icons work on wayfire.

leigh123linux:
This commit doesn't account for the cinnamon panel, the desktop window size needs to minus the panel height.

PR #3388, opened 2024-04-09
nemo-navigation-action.c: duplicate tab by middle-clicking the reload button

ericek111:
I've started using Nemo alongside Caja and realized I miss this feature.

https://github.com/mate-desktop/caja/pull/1456

### DIFF
--- a/src/nemo-navigation-action.c
+++ b/src/nemo-navigation-action.c
@@ -191,8 +191,10 @@ static gboolean
 new_tab_action (NemoNavigationAction *self)
 {
 	NemoWindow *window;
+	NemoWindowSlot *slot;
 
 	window = self->priv->window;
+	slot = nemo_window_get_active_slot (window);
 
 	switch (self->priv->direction) {
 	case NEMO_NAVIGATION_DIRECTION_FORWARD:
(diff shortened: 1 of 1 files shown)


anaximeno:
Should this be made for the home and computer buttons as well?

ericek111:
Oh, I personally don't use those buttons (I have the sidebar always open), but yes, definitely. It should be consistent with the rest of the toolbar.

mtwebster:
This doesn't make any sense - nothing about 'reload' makes me think 'new tab'. It also has zero discoverability beyond looking at the commit history.

I'd consider adding 'New tab' and/or 'New window' toolbar buttons before this.

ericek111:
Why does middle-clicking the back/forward/up buttons open a new card? How would I have figured it out with no manual? Using my common sense and simply trying it (it's literally one click). 

Every major web browser (except Safari) opens a new tab with the same URL upon middle-clicking the reload button. I use it all the time, so it only makes sense I'd expect the same feature from a file manager.

I don't want to have to spin up my own fork just for this, so I guess I'll have to go against my instincts and middle-click the path label of the current directory instead.

Jeremy7701:
A) Why is a new tab function connected to a reload button in a web browser?
B) Since nemo is not a web browser, why would it act like some web browser?
C) Nemo already has the CTRL+T shortcut to do exactly what you want.
D) Alternatively, F3 will give a dual pane view - allowing you to see two panes simultaneously.

ericek111:
A) Ease of use? Instead of the button doing nothing, which isn't very useful, might as well have it open a new tab. Reload button = same location, middle-clicking on a location label = new tab, hence middle-clicking the reload button = open the same location on a new tab.
B) I'm trying to demonstrate that software packages used by billions of people have this feature, so it's safe to assume users won't be confused by it. I mean, it really doesn't have any side effects, it just opens a new tab.
C) Yeah, sure, why do we even have the "up" (go to parent folder) button? You can just erase everything after the last / in the path bar... Again, ease of use. No need to touch your keyboard to duplicate a tab.
D) That's something entirely different.

I really don't see any downside in this. I didn't expect to have such a lengthy discussion about a very trivial change. It's not like this is a major change in UX.

PR #3400, opened 2024-05-12
make nemo build under Ubuntu Noble

pdenapo:
This PR aims to fix two issues detected when building nemo Debian package from the sources using
mk-build-deps  and then fakerrot ./debian/rules binary

1) Update glib version to 2.80 

This is required since the function

[This package cannot be built under Ubuntu jammy any more, where I got the following error message]

nemo-main-application.c:721: reference a `to_log_writer_default_set_debug_domains' undefined

This function is available since: 2.80

https://docs.gtk.org/glib/func.log_writer_default_set_debug_domains.html

2) Remove usr/lib/*/nemo/nemo-extensions-list" from debian/nemo.install

This is required to fix the following error message (since now the nemo extensions are kept in separate packages)

dh_install
dh_install: warning: Cannot find (any matches for) "usr/lib/*/nemo/nemo-extensions-list" (tried in ., debian/tmp)

dh_install: warning: nemo missing files: usr/lib/*/nemo/nemo-extensions-list
dh_install: error: missing files, aborting
make: *** [debian/rules:19: binary] Error 255

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+nemo (6.0.2ubuntu1) virginia; urgency=medium
+
+ * Requires glib >=2.80. 
+ * Remove usr/lib/*/nemo/nemo-extensions-list from nemo.install
+
+ -- Pablo De Nápoli <pdenapo@gmail.com>  Sun, 12 May 2024 14:11:36 +0000
+
 nemo (6.0.2) virginia; urgency=medium
 
   [ Michael Webster ]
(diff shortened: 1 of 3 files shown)


mtwebster:
Hi, nemo shouldn't have any issue building on ubuntu 24.04, as it does have glib 2.80 - maybe you hadn't updated it yet?

I've updated nemo so it will still build on glib < 2.80 installs. This single function isn't worth breaking backward compatibility for folks.

ref: linuxmint/nemo@d1d6bd6b492470143c4f7eac390cd81f3cb84310 

Thanks

PR #3401, opened 2024-05-12
Remove usr/lib/*/nemo/nemo-extensions-list from nemo.install

pdenapo:
This is  a followup of my previous PR #3400 

In order to successfully build the .deb packages for nemo from the sources (using fakeroot ./debian/rules binary)
this other change is requiered, since otherwise one gets the following error message (aborting the building process)

dh_install
dh_install: warning: Cannot find (any matches for) "usr/lib/*/nemo/nemo-extensions-list" (tried in ., debian/tmp)

dh_install: warning: nemo missing files: usr/lib/*/nemo/nemo-extensions-list
dh_install: error: missing files, aborting
make: *** [debian/rules:19: binary] Error 255

This is so, because now the extensions are kept in a separate repository with their own .deb packages

https://github.com/linuxmint/nemo-extensions

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+nemo (6.0.2ubuntu1) virginia; urgency=medium
+ 
+  [ Pablo De Napoli] 
+  * Requires glib >=2.80. 
+  * Remove usr/lib/*/nemo/nemo-extensions-list from nemo.install
+
+ -- Pablo De Napoli <pdenapo@gmail.com>  Sun, 12 May 2024 14:11:36 +0000
+
 nemo (6.0.2) virginia; urgency=medium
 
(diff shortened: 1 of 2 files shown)


mtwebster:
This isn't necessary - I'm not sure why you're having trouble building this, but:

- nemo-extensions-list is *not* an extension, it's a helper for nemo detecting installed extensions. It belongs in the nemo package, and is not optional.
- Please don't modify the debian changelog, this is only done prior to a version tag.

Speaking of the changelog, we aren't affiliated with Ubuntu in any way, as far as this package goes, so we would never use an ubuntu revision. I checked the debian and ubuntu packaging, and neither of them patch any part of the debian packaging here.

PR #3405, opened 2024-05-18
[WIP] Add run in terminal to executable text file preferences

SerenaLynas:
When opening an executable text file, by default it asks whether to run it, including an option to "Run in Terminal". In prefs, you can set what you want so you don't have to specify every time, but curiously "Run in Terminal" is missing from the prefs. This adds it to prefs.

I was not able to get a dev environment set up to run it; the documentation on that seems hidden away and I don't have the time to find it.


gresources/nemo-file-management-properties.glade
```diff
@@ -1234,6 +1234,23 @@ along with .  If not, see <http://www.gnu.org/licenses/>.
                                         <property name="can-focus">False</property>
                                         <property name="orientation">vertical</property>
                                         <property name="spacing">6</property>
+                                        <child>
+                                          <object class="GtkRadioButton" id="scripts_execute_terminal_radiobutton">
+                                            <property name="label" translatable="yes">_Run executable text files in the terminal when they are opened</property>
```
mtwebster, line 1239: You can get rid of the _ in front of Run - this would normally indicate that pressing R should select this option. In reality these shortcuts don't work on this page, so we can leave it out (don't change the other options, as they've already been translated, and any change will void those translations).

gresources/nemo-file-management-properties.glade
```diff
@@ -1234,6 +1234,23 @@ along with .  If not, see <http://www.gnu.org/licenses/>.
...
+                                            <property name="visible">True</property>
+                                            <property name="can-focus">True</property>
+                                            <property name="receives-default">False</property>
+                                            <property name="use-underline">True</property>
+                                            <property name="xalign">0</property>
+                                            <property name="active">True</property>
+                                            <property name="draw-indicator">True</property>
```
mtwebster, line 1246: You need to have:
```
<property name="group">scripts_execute_radiobutton</property>
```

This is in every entry except `scripts_execute_radiobutton` itself, and ties the other buttons together with that one (otherwise you have 2 buttons active at once).

mtwebster:
Hi, this is a bit incomplete:

I've made a couple of inline comments for the glade file.

You also need to make the actual stored settings aware of this new option:
```diff
diff --git a/libnemo-private/org.nemo.gschema.xml b/libnemo-private/org.nemo.gschema.xml
index f54c1dde..1f2d7efc 100644
--- a/libnemo-private/org.nemo.gschema.xml
+++ b/libnemo-private/org.nemo.gschema.xml
@@ -14,6 +14,7 @@
     <value nick="launch" value="0"/>
     <value nick="display" value="1"/>
     <value nick="ask" value="2"/>
+    <value nick="terminal" value="3"/>
   </enum>
 
   <enum id="org.nemo.FolderView">
diff --git a/src/nemo-file-management-properties.c b/src/nemo-file-management-properties.c
index 994e9e66..b9155099 100644
--- a/src/nemo-file-management-properties.c
+++ b/src/nemo-file-management-properties.c
@@ -203,6 +203,7 @@ static const char * const executable_text_values[] = {
        "launch",
        "display",
        "ask",
+    "terminal",
        NULL
 };
```
Otherwise this preference is never even stored.

If you're running Mint, building is as simple as enabling 'source repositories' in your Software Sources in the start menu, then:
```shell
apt build-dep nemo   # installs the sources required to build
cd nemo # your local repo

# build and package
dpkg-buildpackage

#install
cd ..
sudo dpkg -i *.deb
```

mtwebster:
Closing, if you ever update we can re-visit.

PR #3467, opened 2024-10-03
nemo-file.c: "Always" show preview now literally means "always"

### LINKED ISSUE #3291: No image/video preview with MTP connected mobile phone storage (Linux Mint 21.2 - victoria)
### Distribution

Mint 21.2

### Package version

5.8.4+victoria

### Frequency

Always

### Bug description

I am on a fresh installed Linux Mint 21.2 (victoria). If I plug in my mobile phone (Android 10) via USB and activate "file transfer" option on the phone. After this nemo automatically opens the folder where my phone was mounted with mtp://.

After browse to my DCIM folder to look for images i want to copy, no preview will be shown. I have to open every picture what is not very comfortable.
I know that there are some settings under "Edit" > "Preferences" > "Preview" that are correctly set:
image

This works for local, remote or USB storage but NOT for mtp mounted storages.
I have also activated the option "show image preview"
show_imagepreview

By the way, after refresh the folder (Ctrl+R) the checkbox "show image preview" was automatically unchecked.

### Steps to reproduce

plug in a android mobile phone and browse into an image/video folder

### Expected behavior

it should show thumbnails for the stored images or videos

### Additional information

_No response_

jmaebe:
gvfs reports that we should never show thumbnails for files on mtp devices such as phones. Allow users to override this when they select "always". If it gets too slow, the file size filter can still be used to further limit for which files previews are generated.

Add a new "Recommended Only" setting that uses the underlying gvfs recommendation.

Based on https://gitlab.xfce.org/xfce/thunar/-/merge_requests/348

resolves #3291


jmaebe:
One known issue: the added option for showing previews only if recommended by gvfs shifts existing saved options (the options shift one place, unless they were set to "Always"). I'm not familiar enough with gtk to know how to fix this.

mtwebster:
You'd need to update the settings schema also:
https://github.com/linuxmint/nemo/blob/master/libnemo-private/org.nemo.gschema.xml#L2-L6

To keep the original default you'd need to make 'recommended' the first entry I think.

but,

I'm worried this will be confusing to users - what does recommended mean, and how is it different from all? (*I* know, but most users won't)

I'd rather have the 'show thumbnails' (per-folder setting) apply regardless of gvfs settings.

This would complicate `nemo_file_should_show_thumbnail()` a little bit - we wouldn't want to recurse up to the parent for non-local directories. But this makes the behavior opt-in, and permanent for that location.

clefebvre:
I agree with @mtwebster. I don't like this solution but we should fix this issue nonetheless.

Closing this PR for now and adding the issue to the roadmap.

Issue #2494, opened 2020-08-25
Show custom folder/location icons in the sidebar (again)


mscheper:
```
 * Version: 3.8.6
 * Distribution: Mint 19
 * Graphics: Intel Corporation HD Graphics P630
 * 64 bit
 ```

**Issue**

The sidebar currently and exclusively uses what, in #747, are called 'symbolic icons', for the bookmarks and other locations. (I gather these come from the window manager.)

For many users, the custom icons they selected for their bookmarked folders are easier to recognise, especially when they have many bookmarks.

Users should have the option to have their custom icons appear in the Nemo sidebar, as per the regular behaviour in previous versions. Forums abound with requests, confusion and complaints about regression over this issue.

**Steps to reproduce**

1. Go to a folder that contains subfolders.
2. Right-click on the subfolder and select _Properties_.
3. Click on the folder icon, and select a custom icon.
4. Close the _Properties_ dialogue.
5. Drag the subfolder with the custom icon into the sidebar.

**Expected behaviour**

The folder appears with its custom icon in the sidebar.

**Actual behaviour**

The folder appears with the default icon in the sidebar.

**Other information**

It seems a _Sidebar_ Preferences option needs to be added to Nemo, with options for choosing 'symbolic' or regular icons, which sections should be shown, and the background colour, since some custom icons might not look so good on the standard dark grey background. (This can be changed by messing with the style sheet, but that may not be considered good practice.)

landNull:
I would be curious to know the reason for removing this functionality? 
Does it provide any benefit in performance or stability? 
Would adding this functionality back cause any issues?

Jeremy7701:
You need to do it a slightly different way.
1. Select your folder (or file)
2. Right-click on the subfolder and select Properties.
3. Click on the folder/file icon, and select a custom icon (or one of your own design in your home folder).
4. Close the Properties dialogue.

Don't forget that you can use several emblems on icons plus you can design your own and put them in `~/.icons`
AFAIK, the custom icons have to be named with a prefix of 'emblem-'. However, I'm not certain on this point.

For folders, you've also got the colour coding feature...
example:-

![Screenshot from 2021-03-16 13-55-37](https://user-images.githubusercontent.com/52585119/111322254-a1889f80-8660-11eb-9097-07a056d598c9.png)

harrego:
[...]

PR #3516, opened 2025-01-22
add functionality to change bookmark icons

MerQury14092:
I added the "metadata::custom-symbolic-icon-name" attribute to the bookmark files, from which the bookmark icon is formed.

### DIFF
--- a/gresources/nemo-bookmarks-window.glade
+++ b/gresources/nemo-bookmarks-window.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.38.2 -->
+<!-- Generated with glade 3.40.0 -->
 <interface>
   <requires lib="gtk+" version="3.20"/>
   <object class="GtkDialog" id="bookmarks_dialog">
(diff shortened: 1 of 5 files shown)


mtwebster:
Hi, this isn't really something we're interested in. We used to allow colored/custom icons in the sidebar and was deliberately removed. This PR also would also break disconnected state for remote/network bookmarks. Thanks

MerQury14092:
But why did you remove such functionality, it’s convenient when you can customize the icons yourself

theonlywombat:
Frankly, @mtwebster, it sounds like something **you're** not interested in, to the detriment of users.

This feature has been requested repeatedly ever since it was removed, with most receiving plenty of positive attention and support, dating back to at least issue #2494. You've shut down each of these requests and explained that it was "deliberately removed", but you have never explained **why** that is the case.

Even in this thread, you said "This PR also would also break disconnected state for...", with the implication being that you've already stated the reason for its removal and are now providing corollary reasons. So I have to ask: was this the reason for its deliberate removal, or has this reason come up in the time since its removal? If it's the latter, what was the reason for its deliberate removal?

With this history, it's hard for this to not look like you're deliberately dodging the feature for personal reasons. A little clarity goes a long way, and this is one topic that you're unusually unwilling to provide clarity on.

mtwebster:
No, sorry, it was definitely a 'we' decision here.

If you go back to commit histories for various Mint/Cinnamon repos from when this was originally done, you'll probably notice a trend where we're replacing color icons with symbolics *everywhere*, simply because they're much easier to deal with when you start having dark theme variants. A theme only has to define a special 'symbolic' color and symbolic icons will 'just work'.

To clarify, let me rephrase, "Even if we wanted to restore customizable icons to the sidebar, which we don't, this pull request introduces a regression where mount/connected states are no longer identifiable by the item's icon."

https://github.com/linuxmint/xreader/commit/731e015e34c281ea8b742ef7651b991a76b675f8
https://github.com/linuxmint/nemo/commit/4ac714d97fdfa8bd3931b9da95a5867105ebe552
https://github.com/linuxmint/nemo/commit/86927cbfca54fde174a224b5f88ed881c9413dcc
https://github.com/linuxmint/mintupdate/commit/f05327b91b73f3e0ff1b1ba7e7a8fd600ec0462a
https://github.com/linuxmint/mintupdate/commit/d711933f4365bd2aa6065358517b105b0d35e910
https://github.com/linuxmint/nemo/commit/eb0f27d487b110bc1f328fc1a81ce519a88b3985
https://github.com/linuxmint/nemo/commit/2b630d29ce8d23fd5bd9427924a8460a2047714f
https://github.com/linuxmint/nemo-extensions/commit/0abb472236dc3a1d1e87f1085d356094a3bd9f08
https://github.com/linuxmint/cinnamon/commit/61e2c888c69c7827fda5c527d397b46d4f16ca24
https://github.com/linuxmint/mintdrivers/commit/c27e920bb70798b1ac48dbf7ddb59270ed715dee
https://github.com/linuxmint/cinnamon/commit/ef1cf5d99e4e27ad40eb32e1c489812e30a3cdb4
https://github.com/linuxmint/mintsources/commit/e30f82b3a6c570f24d65cc0840d9bb84efeb05cd
https://github.com/linuxmint/xfce-panel/commit/a544958a92034331e2e79d6f7ee0d09d80922344
https://github.com/linuxmint/cinnamon/commit/cbbce7e3526c24f6d06dda86db6ac4f3d94e24b8
... and mention in:
https://blog.linuxmint.com/?p=3557

There are many more commits, I just randomly picked some out.

I hope this waste of my time was worth it for you.

MerQury14092:
@mtwebster, I started writing this commit because I changed my home directory a little, mounted most of the directories to my hdd drive, which is why I stopped displaying the standard bookmark icons that were displayed during the installation of the system
изображение
I would like to simply return the corresponding directory icons, because it annoys me so much that they are now the same, because the first thing the eye looks at when looking for something is an image, also once I had a lot of my own added bookmarks, they all have a directory icon, which I would also like to fix

PR #3525, opened 2025-02-05
nemo-view.c: fix crash after some shortcuts

jknockel:
On `action_follow_symlink_callback()` and
`action_open_containing_folder_callback()` we now confirm that exactly one thing is selected.  Previously, if zero things were selected, we would segfault.


ImAwsumm (approved):

mtwebster:
Hi, how do you reproduce the crash? These two menu items shouldn't be available if there's nothing selected (checked in `real_update_menus`).

PR #3529, opened 2025-02-16
Add background color to the floating status bar

shotgunshine:
Adds a background to the floating status bar for improved readability:

float


mtwebster:
This is caused by a lack of support by your Gtk theme - this isn't the way to address it. I'll add support to our fallback theme info.

PR #3556, opened 2025-05-14
Engrampa integration

Kyuyrii:
I made action files so that Nemo could use Engrampa to compress and extract files, just like it does with file-roller.


Kyuyrii:
> Hi, engrampa MATE's archiving tool so it's not something we need as part of the nemo package.

Do you plan to continue using the outdated file-roller in Mint instead of using Engrampa?

Or are you going to fork the file-roller like Mate did when creating Engrampa?

leigh123linux:
Like all Mate packages, engrampa is barely maintained compared to file-roller.
If you don't like older versions, try using a non LTS distro!

https://github.com/GNOME/file-roller/commits/master/

https://github.com/mate-desktop/engrampa/commits/master/

Kyuyrii:
> Like all Mate packages, engrampa is barely maintained compared to file-roller. If you don't like older versions, try using a non LTS distro!

You clearly didn't understand what I meant.

I didn't mean older packages because it's an LTS, but rather using even older packages just to keep using versions that use GTK themes automatically, which is what Linux Mint does now.

Or do you think that in Ubuntu 24.04 Celluloid is also at version 0.21?

leigh123linux:
> > Like all Mate packages, engrampa is barely maintained compared to file-roller. If you don't like older versions, try using a non LTS distro!
> 
> You clearly didn't understand what I meant.

You failed to mention what you meant

> 
> I didn't mean older packages because it's an LTS, but rather using even older packages just to keep using versions that use GTK themes automatically, which is what Linux Mint does now.
> 

I believe the Mint team has decided to theme the gtk4 apps in a future release.

> Or do you think that in Ubuntu 24.04 Celluloid is also at version 0.21?

I haven't used ubuntu or debian based distros for nearly two decades.

Kyuyrii:
> Do you plan to continue using the outdated file-roller in Mint instead of using Engrampa?
> 
> Or are you going to fork the file-roller like Mate did when creating Engrampa?

I thought it would be clear what I was talking about, since I mentioned forking file-roller and even using Engrampa instead.

If the question was about having more recent packages, then it would probably make more sense for me to talk about creating a version of Linux Mint that was based on the non-LTS versions of Ubuntu.

Kyuyrii:
> I believe the Mint team has decided to theme the gtk4 apps in a future release.

Using themes in GTK4 apps and using themes in Libadwaita apps are different.

For example, the current Celluloid is GTK4 and uses Libadwaita, so many themes look bad on it.

While Transmission GTK is GTK4 but not Libadwaita, I have not had any problems using GTK themes on it so far.

The intention of Libadwaita is to make it harder and worse to use GTK themes, to the point where people stop using and creating GTK themes.

PR #3558, opened 2025-05-16
[WIP] Add a dialog to unmount devices when using Wayland

bigbruno:
The dialog that confirms the unmount of devices that are in use by some program uses resources that are incompatible with wayland and generates segmentation fault. I added a check so that in case you are using wayland, another dialog that is compatible can be created.

Resolves the problem reported in: https://github.com/linuxmint/wayland/issues/134

### DIFF
--- a/src/nemo-places-sidebar.c
+++ b/src/nemo-places-sidebar.c
@@ -26,6 +26,7 @@
 
 #include <gdk/gdkkeysyms.h>
 #include <gtk/gtk.h>
+#include <gdk/gdkwayland.h>
 #include <glib/gi18n.h>
 #include <gio/gio.h>
 #include <math.h>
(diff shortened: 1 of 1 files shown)


cyberknight777:
> I'm not sure this is the way to fix this, or the right place.
> 
> In x11 sessions, the gnome vfs backend will show a busy dialog to prevent unmounting if a drive is in use: image
> 
> This is what we should see in wayland sessions also...

Except it isn't shown and instead crashes as pointed out in the linked issue. So how are we proceeding with this?

bigbruno:
image

With the modification that I implemented, in Wayland a dialog also appears informing what is in use, it is a simpler dialog, it does not detect the program icon, but it is working, while in X11 it continues exactly with the dialog that was already used.

cyberknight777:
are there any updates on this?

mtwebster:
This is being implemented for the Cinnamon session in https://github.com/linuxmint/cinnamon/pull/13145, and will eventually provide this dialog for both x11 and wayland. I'm not sure we'd need this at that point, but I'm leaving it open until that other PR is resolved.

mtwebster:
Addressed in https://github.com/linuxmint/nemo/commit/91037e9e165b1e3f38101bd44fad55ea8226614b

PR #3567, opened 2025-06-08
Add an integrated, feature-rich terminal panel

bigbruno:
An embedded terminal in the file manager is essential for some people, including myself. I’m aware of the plugin that allows adding a terminal to Nemo, but it has usability limitations and is not usable on NixOS. Over the past few weeks, I implemented and, along with some friends, tested a different approach to integrating a terminal into Nemo.

I tried to modify Nemo’s existing code as little as possible, avoiding any additional memory or processing overhead when using the file manager without the terminal enabled, and keeping CPU and memory usage low when the terminal is active.

All terminal settings can be accessed by right-clicking on the terminal, and configurations are saved for future use.

We offer 9 theme options and simple ways to change the font size.

And the most useful part: integration options between the file manager and the terminal. You can choose to synchronize navigation from the file manager to the terminal, from the terminal to the file manager, or both.

Our reference for terminal integration is Dolphin, which had the best integration we knew of. However, with this embedded terminal in Nemo, we’ve added a few more features.

For example, if you type a command in the terminal and then change directories with synchronization enabled, the terminal won’t lose the content you were typing. The content is copied, the cd command is used to change directories, and then the text is pasted back into the terminal using terminal features—this doesn’t interfere with the system’s clipboard functionality.

By default, when accessing an SFTP address in the terminal, you can access files via GVFS Fuse. This is useful, but for DevOps work, it falls short since you can't run commands directly on the server. You have to manually establish an SSH connection, which breaks the sync between the terminal and file manager locations.

In our implementation, when you're in an SFTP address, right-clicking on the terminal gives you the option to connect via SSH, with the ability to maintain folder synchronization. This allows for easy graphical file access and editing, while also letting you run commands in the embedded terminal—like starting or stopping services.

There’s also an option to automatically establish an SSH connection when accessing SFTP addresses.

I hope this Pull Request gets accepted, as it's a feature I would genuinely love to see in all file managers.

image

### DIFF
--- a/gresources/nemo-shell-ui.xml
+++ b/gresources/nemo-shell-ui.xml
@@ -71,6 +71,7 @@
             </menu>
             <menuitem name="Show Hide Menubar" action="Show Hide Menubar"/>
 			<menuitem name="Show Hide Statusbar" action="Show Hide Statusbar"/>
+            <menuitem name="Show Hide Terminal" action="Show Hide Terminal"/>
             <separator/>
 			<menuitem name="Show Hide Extra Pane" action="Show Hide Extra Pane"/>
             <menuitem name="Edit Location" action="Edit Location"/>
--- a/libnemo-private/org.nemo.gschema.xml
+++ b/libnemo-private/org.nemo.gschema.xml
@@ -59,12 +59,27 @@
     <value nick="after-current-tab" value="0"/>
     <value nick="end" value="1"/>
   </enum>
-  
+
   <enum id="org.nemo.SizePrefixes">
     <value value="0" nick="base-10"/>
     <value value="1" nick="base-10-full"/>
     <value value="2" nick="base-2"/>
     <value value="3" nick="base-2-full"/>
+  </enum>
+
+  <enum id="org.nemo.TerminalSyncMode">
+    <value nick="none" value="0"/>
+    <value nick="fm-to-term" value="1"/>
+    <value nick="term-to-fm" value="2"/>
+    <value nick="both" value="3"/>
+  </enum>
+
+  <enum id="org.nemo.TerminalSshAutoConnectMode">
+    <value nick="off" value="0"/>
+    <value nick="sync-both" value="1"/>
+    <value nick="sync-fm-to-term" value="2"/>
+    <value nick="sync-term-to-fm" value="3"/>
+    <value nick="sync-none" value="4"/>
   </enum>
 
   <schema id="org.nemo" path="/org/nemo/" gettext-domain="nemo">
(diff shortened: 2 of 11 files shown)


Jeremy7701:
That "terminal panel" screenshot looks seriously wrong.

Apart from the colour decoration...
The user home directory (`/home/bruno`) should be shown as `~` (not `/`)
cd /home is OK
But `ls` should show all the files and directories in `/home`, not the user home directory.

How will this work, reliably, with split panes and multiple tabs?
Will it cause bugs in any of the nemo plugins - presumably `nemo-terminal` plugin is replaced?

bigbruno:
> A captura de tela do "painel de terminais" parece muito errada.
> 
> Além da decoração colorida... O diretório inicial do usuário ( `/home/bruno`) deve ser mostrado como `~`(não `/`) cd /home está OK Mas `ls`deve mostrar todos os arquivos e diretórios em `/home`, não o diretório inicial do usuário.
> 
> Como isso funcionará de forma confiável com painéis divididos e várias abas? Causará bugs em algum dos plugins do Nemo (presumivelmente, `nemo-terminal`o plugin será substituído)?

Colors and display of ls have nothing to do with the built-in terminal, it's my bash configuration

In split panels and in more tabs, it starts other terminal processes, just like it is done with the terminal plugin

With a terminal built into the nemo code, there is no reason to use a terminal extension

leigh123linux:
> That "terminal panel" screenshot looks seriously wrong.
> 
> Apart from the colour decoration... The user home directory (`/home/bruno`) should be shown as `~` (not `/`) cd /home is OK But `ls` should show all the files and directories in `/home`, not the user home directory.
> 

I never liked the powershell prompts either


> How will this work, reliably, with split panes and multiple tabs? Will it cause bugs in any of the nemo plugins - presumably `nemo-terminal` plugin is replaced?

image

bigbruno:
I updated the code, the features remain the same, but the code quality is much better, easier to maintain and the SSH part is faster.

mtwebster:
Hi,
This is very nicely done but I'm not sure about having to maintain a terminal in addition to a file manager, for such a niche feature, when we have nemo-terminal already.

A few quick observations:
- It doesn't check if split view (F3) is on when you initially turn it on.
- `exit` closes the terminal, and no more terminals will spawn in new tabs after that. Restarting nemo reveals that the toggle has been disabled in Menubar->View.

clefebvre:
I agree with @mtwebster. I like it but the complexity/maintenance worries me more than anything. I'd prefer to see it implemented externally.

Something that hasn't been mentioned also is the fact that we currently rely on gnome-terminal and have no reliable XAPP terminal to rely on yet. If we were to implement a terminal application, that complexity and maintenance would happen there, and here in Nemo we'd simply just embed the library widget.

PR #3587, opened 2025-07-09
Places sidebar, allow users to set custom icons.

Partakithware:
[Screenshot from 2025-07-08 21-51-07](https://ibb.co/1fx0vcVB)

[Screenshot from 2025-07-08 13-38-59](https://ibb.co/CpW0nxFP)

[Screenshot from 2025-07-08 13-39-09](https://ibb.co/zHh2W2v6)

I do not expect this PR to be accepted directly. You may view the new files, and changes via diff. This is more for getting views for consideration of such an update. If you would consider this PR, I can finish removing some of the fluff and no longer used "icon=","g_free(icon)" and finish with the rest as needed.

Opt-in to enable, otherwise defaults to normal nemo standard places icons.

To enable use: gsettings set org.nemo.sidebar-panels.custom-icons use-custom-sidebar-icons true
with the updated nemo gschema.

To disable and return to standard nemo symbolic places icons simply set back to false.

New files, 
/src/nemo-places-sidebar-custom-icons.h
/src/nemo-places-sidebar-custom-icons.c

Updates to: 
org.nemo.gschema.xml
nemo-places-sidebar.c
/src/meson.build
nemo-places-sidebar-ui.xml
nemo-actions.h

I think that's it! Thanks. 

My fork has minor details in the readme/a version 6.4.6 bump that definitely don't need to be merged either. Let me know if you like the thought/idea and you can reject this PR, then I can fix it up proper. Do as you wish.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ Nemo also manages the Cinnamon desktop.
 Since Cinnamon 6.0 (Mint 21.3), users can enhance their own Nemo with Spices named Actions.
 
 Forked Changes: Allows setting of icons in places-sidebar but you must allow it via gsettings.schema
+
+The option appears in the right-click context menu. Enjoy.
 ```
 sudo cp ~/Documents/nemo-master/libnemo-private/org.nemo.gschema.xml /usr/share/glib-2.0/schemas/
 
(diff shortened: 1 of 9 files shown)


mtwebster:
Hi, this functionality was deliberately removed a while back, and is not something we're interested in re-adding at this time.

duplicate of: https://github.com/linuxmint/nemo/pull/3516

Issue #3145, opened 2022-12-28
Cannot use keyboard to copy files into empty folder in list view.


jricardo-um:
```
 * Nemo version 5.2.4
 * Is issue with desktop or windowed nemo? windowed
 * Distribution - Ubuntu 22.04.1 LTS
 * Graphics hardware *and* driver used: no idea
 * 32 or 64 bit: 64-bit
 ```

**Issue**
Cannot paste files into empty folder using *only* the keyboard.

**Steps to reproduce**
1. set the view mode to "List View"
2. expand some subfolders and create any empty folder if there is none already empty
3. you should notice you can't expand it
4. use [Ctrl + C] to copy any file(s)
5. navigate to that empty folder
6. try to paste with [Ctrl + V]
7. files will be copied into the folder containing the empty folder, not inside the empty folder

**Expected behaviour**
The observed behaviour is exactly the expected one: if you use [Ctrl + V] while selecting a expanded folder then the files are copied inside, otherwise they're copied into the parent. The problem is that there's no way to expand empty folders, so you can't paste files inside with the keyboard.

A workaround is to enter said folder. But if you enter the folder and then go back you get all folders collapsed again and that really slows the job if you're trying to organize a lot of files. You can also drag and drop with the mouse, but it's tedious having to do that if you're working with the keyboard most of the time.

**Other information**
While most people sya it's nice for empty folders not to have the arrow and not to be expanded, the fact that they cannot be expanded prevents the user from selecting the inside of the folder and copying files inside it.

As far as I know this was a "visual improvement" a long time ago, but came with this problem. I just want a setting to disable that, please.

Issue #894, opened 2015-05-22
Arrow is not always displayed when a folder is not empty


ghost:
I noticed that in Nemo sometimes the arrow is not always displayed when a folder is not empty in list view. That can confuse you and make you think some folders are empty whereas there not.

This is this arrow I talk about
![screen](https://cloud.githubusercontent.com/assets/2451563/7779549/bfb13402-00d5-11e5-861b-8147ef909168.jpg)

Paviluf:
I know this bug is small but it can confuse you and make you think some folders are empty whereas there not. That can lead you to delete some "empty" folders that are not so empty.

Paviluf:
I need to test it more because with the latest version it don't seem to happen.

Cobinja:
I notice this when copying something into an empty folder. Nemo doesn't seem to notice the content change. This also happens when emptying a folder. The arrow then stays.

Vahan86:
@ghost, is this still an issue for you in Nemo 3.0.6?

JosephMcc:
I can reproduce this on the latest Nemo. The "Size" column will change to indicate there is a change in the number of files but the arrow does not update it's visibility. I don't know if this is a bug in Nemo itself or an issue with GktTreeView's.

dg1727:
This issue is like #508

JosephMcc:
Duplicate https://github.com/linuxmint/nemo/issues/486

mentrelli:
Nemo 3.2.2. still affected by the bug

Issue #616, opened 2014-05-19
Tree/list view - empty folders - redux


HifeFish:
I am not quite happy with the way https://github.com/linuxmint/nemo/issues/344 was resolved. As of now, a folder in tree or list will not have an expand button if and only if it does not contain any currently visible folders or files.
I have two issues with that behaviour:
1. In list view, I used the option to expand an empty folder frequently, to copy/cut and paste files into a (newly created) empty folder without having to change folders; mostly when navigating by keyboard, thus not wanting to rely on drag & drop.
2. On the other hand, the issue in the tree is still the same for any folder without subfolders containing at least one file, still ending up with showing "(empty)" with no actual use, in contrast with the removed expandability in list view. At least as long as you do not choose to display files in the tree.

In my opinion, the ideal behaviour should be as follows:
1. In list view, always show expand-buttons.
2. In tree view hide expand-buttons if and only if there is nothing to be expanded, i.e. if there are no folders in the default setting or if there are no files or folders when files are displayed in tree.

The discriminative behaviour might be harder to implement, but in my opinion the current situation is worse than before.

<s>EDIT: One more issue with the current behaviour, that would also carry over to my suggestion: The expand-button remains hidden after dragging files or folders into an empty folder, unless you refresh, which is easy enough in list view by pressing F5, but can be a hassle in the tree, because you have to collapse and expand the parent folder to do so.
EDIT2: This goes as far as when creating a subfolder of an empty folder and subsequently opening it, the tree still only shows the initial folder without an expand button.</s>
Covered in https://github.com/linuxmint/nemo/issues/486 and https://github.com/linuxmint/nemo/issues/508 (original issue still stands).

icarter09:
@HifeFish is this still an issue?

HifeFish:
The issue still stands in Nemo 3.6.5.

leigh123linux:
3.6.5 is EOL, feel free to reopen if you can reproduce with nemo-4.4.x

Issue #508, opened 2014-01-15
No arrow to expand file tree after pasting file in list view


wunderlich:
I work with an updated Linux Mint 16, Nemo 2.0.8 in list view and recognized the following behaviour:

After pasting a file/folder into an empty folder, no arrow to expand the (no more empty) folder. Here are three pictures which demonstrate this:

folderB is empty:
![nemo-1](https://f.cloud.github.com/assets/4471366/1919267/dca5a3c4-7dd0-11e3-8af0-1b831aa196ba.png)

Paste a file into folderB:
![nemo-2](https://f.cloud.github.com/assets/4471366/1919269/e157a53e-7dd0-11e3-9672-e520cbe18857.png)

No arrow at file tree, even though fileB is no more empty:
![nemo-3](https://f.cloud.github.com/assets/4471366/1919270/e34f709c-7dd0-11e3-8361-2ca371c5fb34.png)

Vahan86:
Checked in Nemo 3.0.6. The issue is still there.

JosephMcc:
Duplicate https://github.com/linuxmint/nemo/issues/486

PR #3588, opened 2025-07-14
[WIP] Add option: Show expanders in folders with no content

jralo7:
When issue #344 (in list mode, hide expanders for empty folders) was implemented, it did not only hide the arrows but also disable expanding them altogether. That caused several problems, including the issues #486, #508, #616, #894 and #3145.

This pull request adds an option to toggle #334's behaviour (optional, disabled by default). Changes include:
- gresources/nemo-file-management-properties.glade - add setting (false by default, keeps old behaviour)
- libnemo-private/nemo-global-preferences.h - declare function and variable for setting
- libnemo-private/nemo-global-preferences.c - add function to retrieve setting
- libnemo-private/org.nemo.gschema.xml - add setting
- src/nemo-file-management-properties.c - add setting
- src/nemo-list-model.c - add behaviour for when setting is true
- src/nemo-list-view.c - add behaviour for when setting is true

Disclaimers:
- I did use AI, but I compiled and tested the changes in my computer (running LM Cinnamon) and reviewed the code (I know basics of C++, enough to understand a bit).

Edits:
- I think a better solution would be to change the icon on empty folders (like a dash instead of a triangle or making the triangle empty) so you can always expand when needed. But that means changes to themes and GTK stuff I don't know.
- After some more time using, I found that removing files on an expanded subfolder may sometimes crash, I'm looking at it. I'll change the status from `Draft` to `Open` once I fix it and no more bugs appear for a while.


jralo7:
### Suggestion

I just thought that maybe we could just change the expander icon on #334's cases (unexpanded empty folders) from a `▶` to a `━`, while allowing to expand them if the user desires so. Expanded folders, regardless if empty or not, would still get a `▼`.

That would keep #334's aesthetic but retain functionality. No extra setting needed.

If the maintainers consider this a better approach I may close this PR and try to do it that way instead.

PR #3621, opened 2025-10-16
Sidebar treeview for better navigation

kdmenk:
I had an idea to improve navigation in nemo by extending its functionality. (similar like Windows Explorer)
Currently, toggling between the Tree View and Places sidebar feels inefficient, so I explored ways to enhance this experience. I kept the idea and the existing two levels of the side pane, but extended the tree view in the next levels showing the file system directories.
The feature van easy being switched on/off by setting the boolean sidebar->use_file_treeview = TRUE;
I also fixed some memory leaks in the existing implementation.
Key navigation similar to Windows Explorer is also possible with the arrow keys when the widget is active.
The TreeView supports lazy loading and update when the underlying filesystem changes.
Switching to display hidden files <ctrl>h filters also the displayed tree view.

Capture_2025-10-17_01-10-17

Capture_2025-10-17_01-55-09

I added this discussion earlier discussion:   https://github.com/orgs/linuxmint/discussions/1169

### DIFF
--- a/src/nemo-places-sidebar.c
+++ b/src/nemo-places-sidebar.c
@@ -5807,12 +5807,6 @@ nemo_places_sidebar_dispose (GObject *object)
 	    sidebar->tree_view = NULL;
 	}
 
-    if(sidebar->hidden_files_changed_id != 0) {
-        g_signal_handler_disconnect (sidebar->window,
-			                         sidebar->hidden_files_changed_id);
-        sidebar->hidden_files_changed_id = 0;
-    }
-
 	g_clear_object (&sidebar->ui_manager);
 	sidebar->ui_manager = NULL;
 
@@ -6003,7 +5997,7 @@ nemo_places_sidebar_set_parent_window (NemoPlacesSidebar *sidebar,
 					  G_CALLBACK (update_places_on_idle),
 					  sidebar);
 
-    if (sidebar->volume_monitor != NULL) sidebar->volume_monitor = g_volume_monitor_get ();
+    if (sidebar->volume_monitor == NULL) sidebar->volume_monitor = g_volume_monitor_get ();
 
 	g_signal_connect_object (sidebar->volume_monitor, "volume_added",
 				 G_CALLBACK (volume_added_callback), sidebar, 0);


kdmenk:
I Started to integrate with nemo-tree-sidebar-model. 
This is the same model implementation like nemo-tree-sidebar currently is using. Thought it would be a good idea to share and use the same code already known.
For usage, nemo-tree-sidebar-model needs to be extended to handle top tree-nodes for organizing the existing tree nodes. I made some progress with the complete implementaion, but still are not finished with the complete task. 
By the way, nemo-tree-sidebar-model is backward compatible to the existing usage (no changes needed in nemo-tree-sidebar!

kdmenk:
I made progress with FMTreeModel (nemo-tree-sidebar-model.*).
It is completly possible to use this implementation also for the TreeView model in nemo-places-sidebar.
The effort for doing this is higher, but we share lot more code and functionality.

To use FMTreeModel I made some extensions:
- A new top organization node without any URL. Needed for "my Computer", "Network".... (needed 2 rows because of the Device)
- Had to implement tho full attribute support for the gtk_tree_model_get and gtk_tree_model_set functions, which are used very much in nemo-places-sidebar. For gtk_tree_model_set there is no GTK interface extension for the model available. I implemented a function fm_tree_model_set which is identical to gtk_tree_model_set for all parameters.
- To make drag&drop work correct, I had to implement the GTK_TYPE_TREE_DRAG_SOURCE interface in FMTreeModel.

The code of nemo-tree-sidebar-model (FMTreeModel) si completly backward compatible. nemo-tree-sidebar can use the new changes without any change.

kdmenk:
I think, I'm done withe the tree view implementation in nemo-places-sidebar.c
PopUp windows are integrated.
For key navigation up, down, right, left work like known from other products. A enter / CR will send URI to the main window and open it.
Drag&Drop is possible into the Tree-View (and inside for Bookmarks as it already was possible)
It is possible to enable this new feature by setting:
sidebar->use_file_treeview = TRUE;
in nemo_places_sidebar_init. Setting it to idebar->use_file_treeview = FALSE; will enable exacly the old implementation! 

It is easy to implement the new feature in a separate file. I decided not to do this, because it doubles the code functionality.

Hope to get some feedback if everything is ok so far.

src/nemo-tree-sidebar-model.c
```diff
@@ -212,52 +246,75 @@ tree_node_unparent (FMTreeModel *model, TreeNode *node)
...
-	g_object_unref (node->file);
-	g_free (node->display_name);
-    g_clear_object (&node->icon);
-    g_clear_object (&node->closed_icon);
-    g_clear_object (&node->open_icon);
+	object_unref_if_not_NULL (node->file); 		node->file=NULL;
+	g_free (node->display_name); 			node->display_name=NULL;
+	object_unref_if_not_NULL (node->icon); 		node->icon = NULL;
+	object_unref_if_not_NULL (node->closed_icon); 	node->closed_icon=NULL;
+	object_unref_if_not_NULL (node->open_icon);	node->open_icon=NULL;
```
mtwebster: As I've mentioned in another pull request, **g_clear_object() already does what you're trying to do here**. That's its reason for existing, to do exactly this here.

mtwebster:
I tried this, a couple of initial observations:

- Tooltips are empty (It's a little blob that follows the pointer around, currently)
- Crashes when expanding some folders
- Sometimes folders with only files in it will show 'empty' and possibly (but not always) crash, sometimes the arrow disappears (I think that's what you want there?
- There's commented code, build conditionals, and some other unnecessary changes that may be due to not fully understanding some aspects of C.
- The sidebar jumps around a lot as you expand rows, sometimes clicking a bookmark will jump you to another part of the sidebar that exposes that folder.

I'm not really digging this, and it's not something I think is widely asked for. I'd probably add bookmarks and XDG locations to the existing treeview sidebar before overhauling the Places sidebar. It already has devices there.

src/nemo-tree-sidebar-model.c
kdmenk: Agree, thats the better code:
	g_clear_object (&node->file);
	g_free (node->display_name); 			node->display_name=NULL;
	g_clear_object (&node->icon);
	g_clear_object (&node->closed_icon);
	g_clear_object (&node->open_icon);
	
	(and object_unref_if_not_NULL is not needed)

kdmenk:
Following replies:

- tooltip   If fixed the empty tooltops for the tree elements (tooltips are not even displayed in the Neme Show Tree View) and I disdn't implement that. The blue empty tooltip point now doesn't show up any more
- Crashes? Do you have checked out my head reales? I do not have chrashes eny more.
- It seams you tested not with the latest checked in Version. Anyho yes I made a small chane, the if a folder shows emty, the expand button will disappear. Expanding a folder which can't be expanded makes no sense. The Nemo NemoDirectory API is a little bit weak in that area.
- I  added comments, but not for me, I added these for others to understand the code faster. If you think from your design perspectives, comments are not needed (and I didn't find any comments) that that's fine for me and I can remove them, even I have another opinion about that.
- I think, you really used not the latest check out version. This was an issue I fixed, because the URI of an file isn't unique in the sidebar any more to identify a node. So the earlier version found an bookmark with the same URI and jumped to this tree node. In the latest release that should have been fixed already. 
- By the way, the existing nemo code can be improved. For example, is the first goto statement since 30 years I have seen in active code, and I programmed C/C++ for more than 20 years. goto isn't really needed... When I started I have had two crashed because of nor deregistered callbacks.

You think that feature is not really needed? So all Windows users use and like it. Not having a real good tree view for me was a show stopper for usage of Linux Mint. That was the reason, I implemented it, and I think I'm not the only person missing that.

Issue #3637, opened 2025-11-10
Nemo is reporting file size wrong


baronleonardo:
### Distribution

Mint 22.2

### Package version

nemo 6.4.5

### Frequency

Always

### Bug description

The file sizes are wrong because of the conversion from bytes to megabytes
you are dividing by 1000 not 1024

This image illustrate the problem
<img width="1653" height="1311" alt="Image" src="https://github.com/user-attachments/assets/bc5e7745-c739-45a3-874d-858cce6b338e" />

This happens both for file properties and in the status bar of nemo.

### Steps to reproduce

1- get any file
2- click it
3- read the size from status bar
4- you can also right click and check the archive properties

### Expected behavior

the size of files should be correct

baronleonardo:
found a solution
`dconf-editor`
`org>nemo>preferences>size-prefixes`

This is set to `base-10`, it should be `base-2`

The problem is `base-10` is the default!!
`base-2` should be the default

PR #3638, opened 2025-11-10
Change default size prefixes from base-10 to base-2

baronleonardo:
related issue #3637


Jeremy7701:
See Preferences, Display
File size = Decimal gives size in kB etc
File size = Binary gives size in kiB etc
No change necessary.

mtwebster:
This is why there's an option to change this... for a majority of users, base-10 is easier to understand.

PR #3648, opened 2025-12-04
nemo: use `Tab` to navigate between splitted panes

### LINKED ISSUE #3647: use `Tab` to navigate between splitted panes
Use `Tab` key beside `F6` to navigate between opened panes

baronleonardo:
Use `Tab` key beside `F6` to navigate between opened panes
closes #3647


kdedev:
I tested it and it is buggy doesn't work consistently

Issue #3085, opened 2022-09-29
[Suggestion] Preview/Metadata panel


Rexadev:
Here is the perfect UI I imagine - 
![image](https://user-images.githubusercontent.com/62152714/192951014-6453af63-c887-4d8d-afc0-e011b57204c6.png)

```
 * Nemo version (nemo --version)
 * Is issue with desktop or windowed nemo?
 * Distribution - Debian Testing Latest
 * Graphics hardware *and* driver used
 * 64 bit
 ```

Jeremy7701:
Why?

------- Original Message -------

Rexadev:
Too look quickly though files.
Preview - Good for Images, video Thumbnails, Audio Album Art.
Metadata - See File Format, Music Sample-BitRate, File Last Edited, File Created
@Jeremy7701

BlabberBlabbing:
I would have opened a new "feature request" but apparently you don't accept them anymore, despite this one doesn't look like it has been addressed.

To the point:

Let's take the "perfect UI" that the OP imagined (though I don't agree on the concept of "perfect" because it depends on the users' needs). On the right you have an "extra pane". Nemo today features an "extra pane" on the right as well. Unlike Windows explorer (and other file managers) though, this extra pane doesn't enable file previews for common file types like media and documents.

Nemo features previews by pressing the spacebar which is good enough for many users. But not for all, especially those that have folders with many media with odd features.
Imagine for example I have a folder with a thousand drone images of a desert. How can I look at them to find the ones where a person is or which are perfectly stabilized?
Can you see that from the thumbnail? NO
Can you see that from the previews? Probably, but you might need to customize the size for each preview and you definitely need more time to do that VS windows previews which just requires you to select the file.
Imagine you want to pick the best picture of a model shoot: in windows you can fastly select with the arrow keys/mouse one pic after the other to check how good they are in the preview pane. In Nemo you can't - fastly.
And so on.

IMO this can (easily?) be implemented adding this feature to the "extra pane" such that, when the feature is enabled, the extra pane shows the spacebar-activated preview automatically (that is, without pressing preview) and resizes it to fill the selected width of the extra pane. But ofc this can be implemented in other ways as well.
Thanks.

Issue #2431, opened 2020-06-29
Feature Request: Add a preview pane to Nemo.


rtevans01:
* Cinnamon version: 4.4.8
 * Distribution:19.3

As the title says, I'm requesting a preview pane in the Nemo file manager. Evidently, it's already built in since [it's visible if you need to upload an image](https://imgur.com/aQgKe08) to a website. However, I can't find an option in the settings to enable it for regular use. 

BTW, I have 2 other feature requests I'd like to make. Should I submit them here instead of spamming the Issues section or submit them separately?

Thanks.

xenopeek:
The dialog in your screenshot is the file chooser provided by the Gtk+ 3 toolkit library. Nemo makes use of the Gtk+ 3.

rtevans01:
I see. Regardless, I think it would be nice to have that feature included in Nemo if possible.

zeeshanali1993:
Also details pane and arrange files by group feature.

xenopeek:
Duplicate of #2122

folkertm:
Let me write something about this wonderful feature. A field report with Fedora and Gnome. I know this feature was or is already implemented in Gnome and Nemo. But I have always had performance problems with it. My use case is simple, I have local folders with photos in different file sizes. The names of the files are autogenerated, as it is the case today with embedded operating systems for digital cameras. And the view or the view change on the camera is more performant, yes, much more performant than on Fedora with Gnome. When switching from one photo to another, it happens that Fedora/Gnome/Nemo gets stuck and does not return to normal operation. A switch to switch on/off the feature may be useful. But more important is a redesign to solve this performance problem.

Dan-cer:
I just left a comment in #2122.

Issue #2122, opened 2019-04-20
Add a Preview pane integrated in Nemo [Feature request]


Tyco72:
For many users it is essential to have a preview pane integrated in the right side of the file manager, to show a large preview for only the currently selected file.
A preview pane allows to use more efficiently the space in the window, showing much more files in the screen, without to fill up the space with previews of files which don't interest. 
It would avoid also the unwanted work in the background of generating/caching previews for all the images contained in the folder (which can be many hundreds of files and of large size).

I guess it would not a too big effort to add a preview pane, at least for photos and videos, because Nemo is already able to generate previews (as icons) for these file types. Consider that MS Window integrates a preview pane in the file explorer since the times of Windows 98. Here a sample pic, of what I mean:

![Preview_Pane_r](https://user-images.githubusercontent.com/47554612/56459326-3231a480-6392-11e9-981d-c5e972daa73e.jpg)

(Issue reported with Nemo 4.0.6 on Linux Mint 19.1 Cinnamon)

folkertm:
I'm not sure @Tyco72 if does that work. The feature was already integrated in older versions of Gnome, e.g. in Fedora 20. But the algorithm had the problem described. The Photos app has the same problem. The higher the number of files and the larger a file the longer Gnome has worked in the background. If I understand you correctly, you can only preview the currently selected file. What if multiple files are selected?

Tyco72:
Hi, thank you. When you select more files it should be not a problem. The preview pane would show the preview only of the last selected file. Anyway only one image at time would be cached.

icarter09:
Many thanks for submitting your feature request for Nemo. Your feature request has been documented and will be reviewed by the Linux Mint team.

For more information on our workflow and feature requests, read https://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/faq.html.

Tyco72:
Thank you very much. I'm looking forward for this feature.

JayBeeDe:
Hi,
I would also apreciate this very nice feature... For pdf files it is possible to use evince preview API. Or libreoffice can handle pdf as well as odt, etc. And what about audio/video files preview ? (too complicated for the moment I can admit it)

pedroolmedo98:
I would be a wonderful feature.

folkertm:
[...]

PR #3650, opened 2025-12-10
[WIP] Preview pane

mtwebster:
ref: #2122, #3085, #2431

The nemo-preview extension isn't really good long-term to maintain, and is broken on some distros

Just targeting image types here for the moment.

- Visibility tied to per-folder metadata
- Toggle is in status bar (along with split pane) - not sure this is the best place - maybe it should be a normal toolbar button?

todo:
- refine appearance
- additional detail rows?  

related discussion:
- https://github.com/orgs/linuxmint/discussions/949

image


CyrilCommando:
Looking good

PR #3675, opened 2026-01-06
nemo-action-manager.c: Fix layout file monitor and action refresh.

mtwebster:
While troubleshooting a different issue, I noticed nemo's action manager fails to refresh in a number of cases.

Make the file monitor change handler handle more cases, and check both files in the callback. Queue up the refreshes for this and changes triggered by directories containing actions also, to prevent excessive work.

### DIFF
--- a/libnemo-private/nemo-action-manager.c
+++ b/libnemo-private/nemo-action-manager.c
@@ -36,6 +36,7 @@ typedef struct {
     GList *actions_directory_list;
     gboolean action_list_dirty;
     gint layout_timestamp;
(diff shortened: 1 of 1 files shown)


PurpleSmurf2:
Does this also apply to updates using mintUpdate?

mtwebster:
No, this is a pull  request. It hasn't been merged yet. Once it is, and a new version is tagged it will go to mintupdate. This isn't really a critical fix for normal users of action layouts. Under normal use everything works fine.

Issue #3687, opened 2026-01-17
Network Sidebar Items Are Not Sorted


My-Random-Thoughts:
### Distribution

Linux Mint 22.3 - Cinnamon 64-bit

### Package version

6.6.3

### Frequency

Always

### Bug description

Following on from the closed feature request #1614 (Oct 2017 - Jan 2020) the network side bar is still not sorted correctly.
Since FRs are closed as per your FAQ, I am opening this as a bug as it's still not fixed.

<img width="213" height="384" alt="Image" src="https://github.com/user-attachments/assets/02f64014-9aef-4664-8343-9842214dc6e8" />

### Steps to reproduce

Have more than one network item listed in the sidebar

### Expected behavior

Network items in the side bar should be sorted alphabetically.

### Additional information

Please fix

nioncode:
I think this can be closed since #3695 was merged

PR #3695, opened 2026-02-01
[Bugfix] Sort network mounts alphabetically

xavierjurado:
This is a fix for #3687, sorting network mounts alphabetically in the sidebar.

I didn't find any contributing guidelines in the repository, so please advice if something isn't quite right.

| Before | After |
|-----------|---------|
| image | image |

### DIFF
--- a/src/nemo-places-sidebar.c
+++ b/src/nemo-places-sidebar.c
@@ -373,6 +373,16 @@ sort_places_func (gconstpointer a,
     return g_utf8_collate (((PlaceInfo *) a)->name, ((PlaceInfo *) b)->name);
 }
 
+static gint
+sort_mounts_func (gconstpointer a,
+                  gconstpointer b)
+{
+    g_autofree gchar *name_a = g_mount_get_name (G_MOUNT (a));
+    g_autofree gchar *name_b = g_mount_get_name (G_MOUNT (b));
+
+    return g_utf8_collate (name_a, name_b);
+}
+
 static PlaceInfo *
 new_place_info (PlaceType place_type,
                 SectionType section_type,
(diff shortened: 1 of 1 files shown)


src/nemo-places-sidebar.c
```diff
@@ -373,6 +373,17 @@ sort_places_func (gconstpointer a,
     return g_utf8_collate (((PlaceInfo *) a)->name, ((PlaceInfo *) b)->name);
 }
 
+static gint
+sort_mounts_func(gconstpointer a,
+                 gconstpointer b)
+{
+    g_autofree gchar *name_a, *name_b = NULL;
```
mtwebster: g_autofree can only be used on one variable at a time - you need to declare name_a and name_b on separate lines, or else `name_b` won't get freed automatically.
```c
g_autofree gchar *name_a = NULL;
g_autofree gchar *name_b = NULL;
```
though they don't necessarily need to initialize to NULL - in a simple function like this I think it's ok to combine it with the assignment below:
```c
g_autofree gchar *name_a = g_mount_get_name (G_MOUNT(a));
g_autofree gchar *name_b = g_mount_get_name (G_MOUNT(b));
```
The important thing is making sure these get *some* sort of value before the function exits, for autofree not to crash.

Also, please keep a space between function name and opening parentheses in all changes:
```c
// good
return g_utf8_collate (name_a, name_b);
// bad
return g_utf8_collate(name_a, name_b);
```

xavierjurado:
@mtwebster on a somewhat related note, is there a list of issues suitable for new contributors? Not necessarily in this repository I guess, just looking for ways to help Mint.

mtwebster:
Unfortunately, we don't have any system in place currently for new contributors. We're trying to get better at staying on top of issues, I'll keep it in mind going forward, and maybe I'll try to start flagging things.

Issue #507, opened 2014-01-15
[Feature Request]Filter Bar Option in Nemo to filter files inside a Folder


anandrkris:
Nemo can add a filter bar to filter files matching a certain pattern. (IIRC, Dolphin (KDE) has this). I am aware that there is a Search option and also a "Select Items Matching" feature that can be used. But in cases where you just need to filter and view files, this could be useful. Search may take some time to search and display results in a large directory. But I believe filter can do it almost instantaneously.

Sample screen shot for reference.

![dolphin-filterbar](https://f.cloud.github.com/assets/4663298/1919189/7d6ef064-7dcf-11e3-8420-bf39c7e19b26.png)

P.S. Now dont ask me to use Dolphin instead. :wink: I love Cinnamon & Nemo. :smile: Just thought this would be a nice addition.

gopalpeddinti:
I would love this feature too! But now that I am posting after an year almost, does this already exist?

dcbishop:
+1

tobiasBora:
Me too I would be really interested by this feature ! Any news ?

alvago:
+1 

If I need to search something on a server via smb.... it takes sooooo long :/

bidinou:
+ 1
Super useful feature which I use all the time :)

Vahan86:
@JosephMcc, can we mark this issue as a feature request?

hiboudev:
I like this feature too, but isn't the ctrl+f bar enough?

ScottWelker:
+1

mtwebster:
I’ve actually been considering replacing the type-to-select with a view filter.

bigbruno:
"I’ve actually been considering replacing the type-to-select with a view filter."

This is a best idea, type-to-select in nemo is very bad, but if use like dolphin filter make this even better then all file managers.

tiago4171:
+1 @bigbruno

bigbruno:
@mtwebster any news about implementation of this feature?

McDulles:
+1 @bigbruno

rafael-neri:
+1 @bigbruno
This feature is super useful.

fladner:
this feature is very interesting

LordGiodai:
@bigbruno 
That's a great idea!

talesam:
Também adoraria a implementação desse recurso, alguma precisão??

bigbruno:
Hello?
Hello?
Is anyone there?
Hello?
[...]

PR #3714, opened 2026-03-02
file selection / interactive search: match anywhere in the filename, not just at the start

### LINKED ISSUE #2660: Support fuzzy search of files and folders by pressing the string in the view
```
 * Nemo version  4.2.3
 * desktop 
 * Distribution - ubuntu 18 
 * no
 *  64 bit
 ```

When I use nemo to search for files (not Ctrl+F), I open a folder, which contains a lot of files and folders as follows: 
- afile
- bfile
 - cfile
 - dfile 
  
 When I press the button to match the file name, I enter file At this time, nemo cannot help me find the file, because it will only match the file starting with the "file" character. I don’t want this, because sometimes I can only remember a few characters. I hope he can match all the names. Is there any way to do it for files and folders containing "file"? Thanks!!!!

### PR DESCRIPTION
**Pull request by nioncode**
Previously, only the start of the filename was matched. Now, the entire file name is searched (case insensitive), e.g. if you have 3 files named "foobar", "foobaz", "foolol", you can simply start typing "lol" and the last file will be selected automatically.

This applies to both, list view and icon view.

Closes #2660.

### DIFF
--- a/libnemo-private/nemo-icon-container.c
+++ b/libnemo-private/nemo-icon-container.c
@@ -3853,8 +3853,7 @@ nemo_icon_container_search_iter (NemoIconContainer *container,
 			continue;
 		}
 
-		if (strncmp (case_normalized_key, case_normalized_name,
-			     strlen (case_normalized_key)) == 0) {
+		if (strstr (case_normalized_name, case_normalized_key) != NULL) {
 			count++;
 		}
 
--- a/src/nemo-list-view.c
+++ b/src/nemo-list-view.c
@@ -191,6 +191,53 @@ G_DEFINE_TYPE (NemoListView, nemo_list_view, NEMO_TYPE_VIEW);
 
 static gint click_policy = NEMO_CLICK_POLICY_SINGLE;
 
+static gboolean
+list_view_search_equal_func (GtkTreeModel *model,
+                             gint search_column,
+                             const gchar *key,
+                             GtkTreeIter *iter,
+                             gpointer user_data)
+{
+    gchar *name = NULL;
+    gchar *normalized_key = NULL;
+    gchar *case_normalized_key = NULL;
+    gchar *normalized_name = NULL;
+    gchar *case_normalized_name = NULL;
+    gboolean is_match = FALSE;
+
+    (void) user_data;
+
+    if (key == NULL || key[0] == '\0') {
+        return TRUE;
+    }
+
+    gtk_tree_model_get (model, iter, search_column, &name, -1);
+    if (name == NULL) {
+        return TRUE;
+    }
+
+    normalized_key = g_utf8_normalize (key, -1, G_NORMALIZE_ALL);
+    normalized_name = g_utf8_normalize (name, -1, G_NORMALIZE_ALL);
+
+    if (normalized_key != NULL && normalized_name != NULL) {
+        case_normalized_key = g_utf8_casefold (normalized_key, -1);
+        case_normalized_name = g_utf8_casefold (normalized_name, -1);
+
+        if (case_normalized_key != NULL && case_normalized_name != NULL) {
+            is_match = strstr (case_normalized_name, case_normalized_key) != NULL;
+        }
+    }
+
+    g_free (case_normalized_name);
+    g_free (normalized_name);
+    g_free (case_normalized_key);
+    g_free (normalized_key);
+    g_free (name);
+
+    /* GtkTreeView expects FALSE for a row that matches. */
+    return !is_match;
+}
+
 static const char * default_trash_visible_columns[] = {
 	"name", "size", "type", "trashed_on", "trash_orig_path", NULL
 };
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by mtwebster** [comment]
This would be a significant change to existing behavior, and I'm not sure it's something   worth having an option for.

**Comment by nioncode** [comment]
Do you see any downsides of the new behavior? It simply helps to find more files that were previously hard to find if you did not exactly know the start of the filename. The linked issue is 5 years old, so it seems like at least some users wanted to have this feature for a long time and it would be great if this can either be merged as-is or with an additional option if you don't want to break existing behavior.

**Comment by mtwebster** [comment]
> Do you see any downsides of the new behavior?

Yes - it's completely different from the old behavior.

This is used but most people to immediately jump to a file, or at least get close alphabetically. With this change there could be 100s of files with matches I'd have to wade thru before getting where I want. With that extra effort it's just as easy to do a normal search (<kbd>ctrl</kbd>+<kbd>F</kbd>) instead.

Some = two users: The OP and yourself. I don't see the need for it to be an option, just use search. It'll probably be quicker in the long run.

**Comment by nioncode** [comment]
I'll just describe one of my most relevant use cases here: navigating a directory tree. Usually you pretty much know some mostly unique sequence of chars in your filenames. However, if you just support selecting the folder by typing the start of the file name, this does not help you at all if you have multiple folders starting with the same prefix.

I don't know how search would help here, since it descends into subdirs and just spams my output with hundreds of files deeply nested, which has nothing to do with the files in the current folder that I actually want to reach (not even mentioning that you need to ctrl+f, tabbing around in the interface etc., which takes forever compared to the quick selection).

If you still think this change is not relevant, not even as an optional behavior which I'd happily add, then I'll move on and find another file manager.

**Comment by Jeremy7701** [comment]
**I'm mystified.**
As far as I can tell search in nemo 6.6.3 searches for a match **anywhere** in a filename
So a search for 'ps'  in the /etc folder matches (recursive search):-
```
[many lines omitted]...
/etc/apparmor.d/local/usr.sbin.cupsd
/etc/xdg/menus/applications-merged/webapps.menu
/etc/menu-methods/xdg-desktop-entry-spec-apps
/etc/xdg/xdg-default/autostart/xfce4-tips-autostart.desktop
```
I call this a match anywhere in the filename, not just at the start, as two people have requested.
Since I almost always use the initial few characters of filenames, this doesn't make a big difference for me.

nioncode:
I'm not disputing that search does not work correctly. I'm just saying it is the completely wrong tool to use in my use case. If I hit ctrl+f and search for something in my home, this will throw thousands of files at me and will bury the actual file that I want to select in a huge pile of other files (that are deeply nested in subdirectories), which makes it even harder to select than just selecting it manually from the current directory.

alexeymuranov:
> I'm mystified.

@Jeremy7701, i believe to understand that this is not about the search through the "Search for files" menu item. It is about the automatic file selection when you start typing.

Altogether, I think there is a reason that fuzzy finding extensions are so popular. I am using [fzf](https://github.com/junegunn/fzf).

So, maybe this could be implemented as a Nemo plugin?

KonTy:
> > I'm mystified.
> 
> @Jeremy7701, i believe to understand that this is not about the search through the "Search for files" menu item. It is about the automatic file selection when you start typing.
> 
> Altogether, I think there is a reason that fuzzy finding extensions are so popular. I am using [fzf](https://github.com/junegunn/fzf).
> 
> So, maybe this could be implemented as a Nemo plugin?

This could be a setting, I've been piloting these features forawhile in my private fork, if you want a version that does that check here: https://github.com/smpl-os/nemo-smpl

it has a host of new features that I added for myself including fuzzy finder, Preview with Alt-F3 and a lot more.

mtwebster:
> I don't know how search would help here, since it descends into subdirs and just spams my output with hundreds of files deeply nested

You can turn off recursion in the search bar.

-----
This reminded me of another feature request from even longer ago, that would maybe be a better fit - this sort of matching would work better if the whole 'type-to-search' feature worked differently - having the view filter out non-matching files as you type.

It was originally brought up in https://github.com/linuxmint/nemo/issues/507

Similar to a normal non-recursive file search, but sorted by match quality (prefix matches would count higher than inline ones). This could make it palatable to people who prefer the current behavior, since a direct match would still be first in the results.

I resurrected a branch I'd started a couple of years ago:
image

Jeremy7701:
You might be interested in [https://en.wikipedia.org/wiki/Soundex](https://en.wikipedia.org/wiki/Soundex). Not totally reliable for many filenames, though!

alexeymuranov:
@mtwebster 

> With that extra effort it's just as easy to do a normal search (ctrl+F) instead.

> You can turn off recursion in the search bar.

Searching with <kbd>Ctrl+F</kbd> is a considerable extra effort:

 1. First, I need to type <kbd>Ctrl+F</kbd>. Typing this combination takes about the same amount of time/effort as typing 3 letters of the file name, which could be already enough to focus to the desired file.

 2. Then, I need to turn off recursion. How to do this without using mouse? Then, when I do a normal file search, I need to turn it back on.

 3. In the end, I need to hit <kbd>Enter</kbd>. I also do not see interactively if I've typed enough.

 4. Then what? How do I open the found folder without using the mouse?

The current "auto-focus" behaviour, that focuses on the name prefix, is not good for me because I have many folders with the same technical prefix.

PR #3716, opened 2026-03-04
add 'copy path' to context menus

kavehtehrani:
Adds a "Copy Path" item to the right-click context menu that copies the full path to the clipboard. For multiple selections, paths are on a line each. Falls back to URI for nonlocal files (smb://, ftp://, etc). Toggleable in Preferences / Context Menus. 

I followed the pattern for 'copy' more or less. One area I'm not sure, is the sensitivity check for NEMO_ACTION_COPY_PATH in real_update_menus(), because it's technically redundant since the selection menu only appears with a selection, but I followed the existing pattern (including 'copy') which use clipboard. Hope I did this right but pls let me know otherwise.

### DIFF
--- a/gresources/nemo-directory-view-ui.xml
+++ b/gresources/nemo-directory-view-ui.xml
@@ -174,6 +174,7 @@
 	<placeholder name="File Clipboard Actions">
 		<menuitem name="Cut" action="Cut"/>
 		<menuitem name="Copy" action="Copy"/>
+		<menuitem name="Copy Path" action="Copy Path"/>
 		<menuitem name="Paste Files Into" action="Paste Files Into"/>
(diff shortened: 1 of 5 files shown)


mtwebster:
https://github.com/linuxmint/nemo/pull/3719#issuecomment-4009161337

kavehtehrani:
That's cool. Didn't know it already existed but it's kind of hidden. Think it would be worth it to make it a bit more explicit but yeah pasting into text boxes does work!

mtwebster:
True it's technically hidden, but it's really just the normal way the system clipboard works in general.

For the most part, unless it's some new core functionality, extra context menu items should  be added as NemoActions, via spices if you want them available to others. As I said, there is already one for this (though I don't know why or if it falls back to urls the same way).

PR #3718, opened 2026-03-06
feat: match anywhere in filename during interactive search

KonTy:
## Summary

Changes the interactive type-ahead search (the search that activates when you start typing in a file view) to match **anywhere** in the filename instead of only matching from the beginning.

### Changes

- **Icon view** (`nemo-icon-container.c`): Changed `strncmp(key, name, strlen(key))` → `strstr(name, key)` so typing e.g. "log" matches "changelog.txt"
- **List view** (`nemo-list-view.c`): Added a custom `search_equal_func` that normalizes both the search key and filename to UTF-8 NFD + casefold, then uses `strstr` for substring matching

### Motivation

The prefix-only matching behavior is unintuitive — users often know part of a filename but not the exact beginning. Substring matching is the standard in most modern file managers.

Inspired by https://github.com/linuxmint/nemo/pull/3714

### DIFF
--- a/libnemo-private/nemo-icon-container.c
+++ b/libnemo-private/nemo-icon-container.c
@@ -3875,8 +3875,7 @@ nemo_icon_container_search_iter (NemoIconContainer *container,
 			continue;
 		}
 
-		if (strncmp (case_normalized_key, case_normalized_name,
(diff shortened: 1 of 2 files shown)


mtwebster:
apparent duplicate of #3714?

PR #3719, opened 2026-03-06
feat: add 'Copy Path' to context menu

KonTy:
## Summary

Adds a new **Copy Path** item to the right-click context menu that copies the full filesystem path of the selected file(s) to the clipboard. For non-local files (e.g. `sftp://`), the URI is copied instead.

### Changes

- **`nemo-actions.h`**: Added `NEMO_ACTION_COPY_PATH` define and configurable menu entry
- **`nemo-view.c`**: Added `action_copy_path_callback` that builds a newline-separated list of file paths and copies to clipboard via `gtk_clipboard_set_text`; action entry with label "Copy Pat_h"; sensitivity follows `can_copy_files`
- **`nemo-directory-view-ui.xml`**: Added `Copy Path` menuitem after `Copy` in the context menu
- **`org.nemo.gschema.xml`**: Added `selection-menu-copy-path` boolean key (default: true)
- **`nemo-file-management-properties.glade`**: Added "Copy Path" checkbox in the Context Menu preferences tab

### Behavior

- Single file selected → copies its absolute path
- Multiple files selected → copies paths separated by newlines
- Configurable: can be hidden via Preferences → Context Menu

Inspired by https://github.com/linuxmint/nemo/pull/3716

### DIFF
--- a/gresources/nemo-directory-view-ui.xml
+++ b/gresources/nemo-directory-view-ui.xml
@@ -174,6 +174,7 @@
 	<placeholder name="File Clipboard Actions">
 		<menuitem name="Cut" action="Cut"/>
 		<menuitem name="Copy" action="Copy"/>
+		<menuitem name="Copy Path" action="Copy Path"/>
 		<menuitem name="Paste Files Into" action="Paste Files Into"/>
         <menuitem name="Duplicate" action="Duplicate"/>
 	</placeholder>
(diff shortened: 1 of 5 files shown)


mtwebster:
This already works... if you copy some files in nemo, and paste into somewhere that accepts text (instead of a file), it pastes their paths, one per line.

If that wasn't enough there's a Nemo Action that does this also.

PR #3722, opened 2026-03-06
Add configurable keyboard shortcuts

KonTy:
All keyboard shortcuts are now stored in GSettings under `org.nemo.keybindings` and can be changed at runtime. A new **Keyboard Shortcuts** tab in Edit → Preferences lets users view, search, and rebind any shortcut using an inline key-capture widget. Changes take effect immediately without restart.

Additional configurable bindings included:
- **open-default** — open with default application (unbound by default)
- **new-folder-alt** — alternate new-folder key (unbound by default)
- **trash-alt** — alternate move-to-trash key (unbound by default)

These allow distro or user profiles to add secondary bindings (e.g. F4, Insert, Delete) without changing Nemo's stock defaults.

key bindings

### DIFF
--- a/gresources/nemo-file-management-properties.glade
+++ b/gresources/nemo-file-management-properties.glade
@@ -4383,6 +4383,22 @@ along with .  If not, see <http://www.gnu.org/licenses/>.
                     <property name="position">8</property>
                   </packing>
                 </child>
+                <child>
+                  <object class="GtkBox" id="keybindings_box">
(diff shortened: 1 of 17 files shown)


mtwebster:
Thanks, but this is overkill - we have no other individual apps that allow customizing built-in shortcuts. We try to be careful about overlap with default system shortcuts - this will only open folks up to potential conflicts.

PR #3726, opened 2026-03-06
file-operations: throttle page cache during cross-device copies

### LINKED ISSUE #3710: Transfer to USB drive locks up with grouped graphical cut/paste
### Distribution

Mint 22.0 and Mint 22.3

### Package version

6.2.9

### Graphics hardware in use

_No response_

### Frequency

Quite often

### Bug description

When selecting a large number (20) .iso files (3-6GB each) and using graphical copy/paste to a USB3 thumb drive the USB port locks up.  I was copying the .iso files from my NAS drive (Synology) to the USB drive (Ventoy). After about 6-8 files (20 GB worth), all activity stops.  The drive cannot be ejected, (Always busy) and does not recover. Removing the drive or re-inserting it has no effect until a full hard reset. The machine cannot be shut down. Shutdown or Restart attempts immediately circle back to the login screen. After a hard power off reset, I tried to copy the remaining files resulting in another lockup. Five of the .iso files were corrupted (sha256sum mismatch)

I manually deleted and re-copied the corrupted files one at a time via the terminal without any problems.  I cannot determine if this is a GUI (Cinnamon) issue or deeper in the USB driver.
This error is occurring on two widely different machines and two very different USB drives. (128-GB size)
- Dell XPS 7590 laptop (Mint20.0)
- White box ASROCK Ryzen-5 system (Mint20.3)  with the same lockup and unable to shut down result.
- Drives: Team Group medium speed USB3 (60MB/s) and SSK high speed USB3 (200MB/s) 

### Steps to reproduce

Input:
In source GUI window, group select 15-20 3-5GB binary files, right click, copy.
Switch to thumb drive GUI window, right click, paste.

Result:
Locks up after 5-8 files have been copied.
Transfers stop. CPU idles, did not resolve after waiting 15 minutes.

### Expected behavior

Expect files to be copied without errors or corruption.

### Additional information

_No response_

KonTy:
Fixes #3710

When copying many large files (15-20 ISOs, 3-6 GB each) from a fast source (NAS/SSD) to a slow USB drive, GLib's splice() fills the page cache at RAM speed. Once dirty pages hit the kernel's dirty_ratio, all writes block and the system freezes. I've seen the same behavior in Double Commander on Arch, where I have to enable "verify on copy" to make sure files aren't silently corrupted. The fix adds periodic sync_file_range() + posix_fadvise(DONTNEED) calls every 32 MiB in the copy progress callback, which provides natural back-pressure that paces the copy to actual device speed and keeps dirty pages bounded at ~32-64 MiB instead of many gigabytes.

### DIFF
--- a/libnemo-private/nemo-file-operations.c
+++ b/libnemo-private/nemo-file-operations.c
@@ -26,6 +26,10 @@
             Pavel Cisler <pavel@eazel.com>
  */
 
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE  /* for sync_file_range() */
+#endif
+
 #include <config.h>
 #include <string.h>
 #include <stdio.h>
@@ -34,6 +38,7 @@
 #include <math.h>
 #include <unistd.h>
 #include <sys/types.h>
+#include <fcntl.h>
 #include <stdlib.h>
 #include <errno.h>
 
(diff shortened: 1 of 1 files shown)


Jeremy7701:
This "problem" has never occurred for me.
If you believe it is a kernel problem, then you should contact the kernel maintainers.
User workarounds won't solve any genuine problems.

PR #3728, opened 2026-03-07
Use cover.{jpg,png} as directory icons

WhyNotHugo:
Use cover.jpg or cover.png as icons for directories (unless anything else is explicitly configured).

This is a relatively common convention supported by various tools, and especially convenient for things like music libraries.

### DIFF
--- a/libnemo-private/nemo-file.c
+++ b/libnemo-private/nemo-file.c
@@ -4343,6 +4343,25 @@ get_custom_icon (NemoFile *file)
 		icon = g_object_ref (file->details->custom_icon);
  	}
 
+	if (icon == NULL && nemo_file_is_directory (file)) {
+		static const char *cover_names[] = { "cover.jpg", "cover.png", NULL };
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
Gentle ping?

mtwebster:
This isn't really the way to go about this - the changes here can block while checking if the file exists.

Have you tried https://github.com/flozz/cover-thumbnailer?

This actually sets the folder's metadata to use either a generated or existing thumbnail file provided in the folder. This allows these images to be available asynchronously, and part of the normal thumbnailing process.

WhyNotHugo:
> Have you tried https://github.com/flozz/cover-thumbnailer?

I think this relies on gvfs and the rest of the GNOME stack, right?

I'm not running nemo under cinnamon or GNOME, so that approach is not really feasible for me.

WhyNotHugo:
Would  a non-blocking implementation be acceptable, or is my usage unsupported?

PR #3741, opened 2026-03-21
Add custom date format and alternating row colors in list view

StefanL38:
## Summary

This PR adds two new features to the Nemo list view:

### 1. Custom Date Format
A new \"Custom…\" option in the date format dropdown under
*Edit → Preferences → Display*.

When selected, a text field appears where the user can enter a
format string using the following tokens:

| Token | Meaning           | Example |
|-------|-------------------|---------|
| YYYY  | 4-digit year      | 2026    |
| YY    | 2-digit year      | 26      |
| MM    | Month 01–12       | 03      |
| DD    | Day 01–31         | 21      |
| HH    | Hour 00–23 (24h)  | 14      |
| hh    | Hour 01–12 (12h)  | 02      |
| mm    | Minute 00–59      | 35      |
| SS    | Second 00–59      | 07      |

Example: `YYYY.MM.DD HH:mm` → `2026.03.21 14:35`

Changes apply immediately without restarting Nemo.

Two additional buttons allow managing date column widths:
- **Auto-fit**: sets all date column widths to text width plus a saved padding value
- **Save padding**: saves the current padding as the new default for Auto-fit

### 2. Alternating Row Colors
A new checkbox *"Alternating row colors (white / light gray)"* under
*Edit → Preferences → Views → List View Defaults*.

When enabled, odd rows are displayed with a light gray background (`#ebebeb`) and even rows with the default background. This improves readability in long file lists.

Enabled by default.

**Implementation note:** CSS `nth-child` selectors are not reliable for `GtkTreeView` rows in GTK3 (row nodes are recycled dynamically). The alternating background is therefore implemented via `GtkCellRenderer::cell-background` using cell data functions, which is the only robust approach in GTK3.

## Changed files
- `libnemo-private/org.nemo.gschema.xml` – new keys: `date-format-custom`, `date-column-autosize`, `date-column-padding`, `list-view-alternating-rows`
- `libnemo-private/nemo-global-preferences.h/.c` – new preference constants and cached format string
- `libnemo-private/nemo-file.h/.c` – `nemo_custom_date_format_to_strftime()`, `nemo_file_get_sample_date_string()`, `NEMO_DATE_FORMAT_CUSTOM` case
- `src/nemo-file-management-properties.c` – custom format entry widget, Auto-fit and Save padding buttons, combo change handler
- `src/nemo-list-view.c` – `set_row_background()`, `icon_cell_data_func()`, `text_cell_data_func()` for alternating rows
- `gresources/nemo-file-management-properties.glade` – new widgets for custom format box, buttons, and alternating rows checkbox
- `po/de_custom.po` – German translations for all new UI strings

### DIFF
--- a/gresources/nemo-file-management-properties.glade
+++ b/gresources/nemo-file-management-properties.glade
@@ -938,14 +938,14 @@ along with .  If not, see <http://www.gnu.org/licenses/>.
                                         </child>
                                         <child>
                                           <object class="GtkCheckButton" id="list_view_alternating_rows_checkbutton">
-                                            <property name="label" translatable="yes">Abwechselnde Zeilenfarben (weiß / hellgrau)</property>
+                                            <property name="label" translatable="yes">Alternating row colors (white / light gray)</property>
                                             <property name="visible">True</property>
                                             <property name="can-focus">True</property>
                                             <property name="receives-default">False</property>
                                             <property name="use-underline">True</property>
                                             <property name="xalign">0</property>
                                             <property name="draw-indicator">True</property>
-                                            <property name="tooltip-text" translatable="yes">Zeilen abwechselnd weiß und hellgrau darstellen, um die Lesbarkeit zu verbessern.</property>
+                                            <property name="tooltip-text" translatable="yes">Display rows alternating white and light gray to improve readability.</property>
                                           </object>
                                           <packing>
                                             <property name="expand">False</property>
--- a/gresources/nemo-file-management-properties.glade
+++ b/gresources/nemo-file-management-properties.glade
@@ -936,6 +936,23 @@ along with .  If not, see <http://www.gnu.org/licenses/>.
                                             <property name="position">1</property>
                                           </packing>
                                         </child>
+                                        <child>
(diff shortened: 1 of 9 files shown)


mtwebster:
Hi, thanks for the submission, but I think we'll pass on this.

General observations:
- PRs should be tightly focused on a single improvement, fix or feature. There are at least two to three unrelated sets of changes here.
- Don't update translations - that gets done via Launchpad imports. Anything changed here will get overwritten before the next major release.

Specific observations:

#### Zebra striping
This should be left up to the Gtk theme, not hardcoded. This is just adding complexity to the rendering of the view. This is stated in the documentation (for the 'rules-hint' property, which used to provide a programmatic method you're re-implementing here.

#### Custom date
This is overkill here - if there is another, specific useful format we're not covering in the list, we could add it.

It's definitely not worth the added complexity here in the list view (which I'm not sure is necessary in the first place - the column should size to fit).

JosephMcc:
> This should be left up to the Gtk theme, not hardcoded

Just a quick note. I don't think this has been supported in Gtk themes for quite awhile.

mtwebster:
Either way, I don't think it's something that needs to be resurrected.

PR #3746, opened 2026-03-26
Implement a filtered view for interactive-search.

mtwebster:
This would replace the current interactive search feature (typing to restrict selection to filenames beginning with the typed text), except for the desktop view.

Instead, typing would show an entry as before, but now:

- The view will actively filter out non-matching files.
- Matches will by made and scored using the 'fzy' matching algorithm based on https://github.com/jhawthorn/fzy.
- Matching characters will be highlighted (bold) within the filename in the resulting list.
- Focus/navigation remains in the view so keyboard navigation can still be performed while the filter is active.
- Deleting (backspace) all text from the filter hides the filter and restores the normal view.

ref: #3714, https://github.com/linuxmint/nemo/issues/507

### DIFF
--- a/config.h.meson.in
+++ b/config.h.meson.in
@@ -17,8 +17,6 @@
 #mesondefine HAVE_EXIF
 // Define if libselinux is available
 #mesondefine HAVE_SELINUX
-// Define to enable pango-1.44 fixes
-#mesondefine HAVE_PANGO_144
 // Define to use gtk-layer-shell
 #mesondefine HAVE_GTK_LAYER_SHELL
 
--- a/debian/control
+++ b/debian/control
@@ -23,7 +23,7 @@ Build-Depends:
  libgtk-3-doc,
  libgtk-layer-shell-dev,
  libjson-glib-dev (>= 1.6),
- libpango1.0-dev,
+ libpango1.0-dev (>= 1.44.0),
  libx11-dev,
  libxapp-dev (>= 2.0.0),
  libxext-dev,
--- a/eel/eel-editable-label.c
+++ b/eel/eel-editable-label.c
@@ -998,9 +998,7 @@ eel_editable_label_ensure_layout (EelEditableLabel *label,
       if (label->font_desc != NULL)
 	pango_layout_set_font_description (label->layout, label->font_desc);
       
-#ifdef HAVE_PANGO_144
       pango_attr_list_insert (tmp_attrs, pango_attr_insert_hyphens_new (FALSE));
-#endif    
       pango_layout_set_attributes (label->layout, tmp_attrs);
       
       if (preedit_string)
--- a/libnemo-private/fzy-bonus.h
+++ b/libnemo-private/fzy-bonus.h
@@ -1,0 +1,132 @@
+/*
+ * fzy fuzzy matching algorithm - bonus scoring tables
+ *
+ * Copyright (c) 2014 John Hawthorn
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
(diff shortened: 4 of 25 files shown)


Jeremy7701:
1. If I want to search using a partial file name, will it be possible to do this without an **algorithm intruding** into the search and sorting the output?
2. Will a recursive search be possible?
3. Will a case sensitive search be possible?
The current search locates the search string anywhere within the filename - not just matching from the first character.
Why is a change necessary?

mtwebster:
This is to replace the current interactive search for the current view, it has nothing to do with the real search/content search (control-F), and only applies to the current view (non-recursive). It's to aid *selection* not finding - you know the file probably exists.

The current interactive search ('just start typing') only matches starting characters in a filename. #3714 aimed to make that match anywhere in a filename, which I felt is more of a hindrance than an improvement, since it can result in meaningless matches you need to navigate past to get to the file you wanted.

Jeremy7701:
**I wasn't aware of this file positioning function!**

Obviously I have never used it, and for me, the old version is sufficient.
Please ignore all my previous comments as they aren't relevant.

mtwebster:
merged: https://github.com/linuxmint/nemo/commit/11df68ccb3ca1e3787959b1d36b75c10bd4d339c

nioncode:
Sorry, I only got around to testing this now. The functionality in generally works fine, thank you :)

However, it does not really solve my original use case (neither does my proposed #3714): navigating a directory tree fast. I'm not sure whether I should open a separate issue for this or where to discuss this best, please let me know the best place to put this (since this closed MR is certainly not the correct place).

Let's assume the following directory tree:

```
.
├── bar
└── foo
    ├── aa
    └── baz
        ├── bb
        └── target
```

What I currently have to do with this feature:
* type 'fo' to filter
* hit 'arrow down' to select the `foo` folder
* hit 'enter' to enter the directory
* type 'ba' to filter
* hit 'arrow down' to select the `baz` folder
* hit 'enter' to enter the directory
* type 'ta' to filter
* hit 'arrow down' to select the `target` folder
* hit 'enter' to enter the directory

I'd like to get rid of having to hit 'arrow down' to select the directory, since this requires moving my hand to the arrow block on the keyboard and then back to the 'main area' (don't know how to describe it best). On my old file browser (spacefm), the first entry in the view is always selected by default (although it is a kind of 'soft selection', i.e. it is not completely highlighted, but only drawn with a border; I don't know what the difference is between those states though), so you can just hit enter directly whenever you enter a directory and you want to go into the first folder (or the first match after typing).

Do you think it makes sense to add a similar behavior to nemo, so that you can:
- 1) navigate a deep directory tree just by hitting 'enter' multiple times (as long as you just want to go into the first directory, which is common for many java repositories that e.g. start with `src/main/com/github/package`
- 2) directly enter a directory after having filtered the view to a match without having to hit 'arrow down'

mtwebster:
@nioncode I just made a change to the filter view - it will automatically select the first item when the text changes.

https://github.com/linuxmint/nemo/commit/bc0c8cc8e252701dccd7a5e4e61b7fbf7b838a7b

Jeremy7701:
Will some detailed documentation be available when the new filter option is released?

mtwebster:
I don't think that's necessary, there's not much to it.

nioncode:
@mtwebster your commit looks great!

Would it also make sense to always select the first file/directory by default (independent from this interactive search feature) so that hitting enter repeatedly allows to traverse a tree of `src/main/java/com/github/nioncode/my-project/` simply by hitting enter a few times without needing to hit arrow down in each directory?

nioncode:
@mtwebster not sure if you missed my comment or not. If you are open to selecting the first file/dir by default, I'd be happy to submit a PR for that

mtwebster:
I'm not convinced of doing that in *all* situations.  It does select the first 'result' automatically now when engaging the filter, though.

nioncode:
Do you have any concrete ideas where selecting the first file might be bad? I have many deeply nested directories with only one dir inside where it would be very beneficial to simply hit `Enter` a few times on the keyboard to traverse them.

Tanza3D:
Can we have an option to go back to the old style? This new one is incredibly frustrating to me as someone who's used the classic style of typeahead search since forever - I switched to nemo pretty much exclusively *beacuse* it had this style of typeahead search, it's quite disappointing to lose it all of a sudden now.

PR #3748, opened 2026-03-29
a proper drop shadow for nemo-desktop

perdakovich:
(hover over images to see their values) a few presets could be presented in GUI settings

i do this via `~/.config/gtk-3.0/gtk.css` :
```
/* Drop shadow for Nemo-desktop (Cinnamon) */
.nemo-desktop.nemo-canvas-item {
    text-shadow: 2px 2px 3px alpha(black, 0.8);
}
```

i had this tweak for some years and i think the time to bring it up has come.

png-bob

everyone deserves some nice and clean drop shadows

### DIFF
--- a/gresources/nemo-style-application.css
+++ b/gresources/nemo-style-application.css
@@ -24,8 +24,8 @@ NemoDesktopWindow GtkPaned {
 }
 
 .nemo-desktop.nemo-canvas-item {
-    color: #eeeeee;
-    text-shadow: 1px 1px alpha(black, 0.8);
+	color: #eeeeee;
+    text-shadow: 2px 2px 3px alpha(black, 0.8);
 }
 
 .nemo-desktop.nemo-canvas-item:hover {


JosephMcc:
I wouldn't call that nice and clean. I'd call it way over done personally.

perdakovich:
okay. everyone deserves drop shadows that suits their taste, right? 
then i propose making a way to choose them. perhaps a few presets somewhere in settings could be enough, for more advanced customization there's always `gtk.css`
that will involve more complex fiddling with gui interfaces, so for that one, i'm just the ideas guy

Jeremy7701:
What does it look like on dark themes?

perdakovich:
what do you mean?

mtwebster:
We're working on making this a bit more flexible - https://github.com/linuxmint/nemo/pull/3754 so changing the current value doesn't make sense at this point.

PR #3749, opened 2026-03-30
Add NemoSelectionProvider extension interface

HollowPhobos:
Adds a new NemoSelectionProvider interface to libnemo-extension that allows extensions to be notified when the file selection changes.

- Add nemo-selection-provider.h/.c with the new interface
- Call nemo_view_notify_selection_providers() from nemo_view_notify_selection_changed() to notify all registered selection provider extensions


HollowPhobos:
closed for being ignored

Issue #3750, opened 2026-03-31
Selecting "rename" in file conflict dialog overwrites parent directory if directory scan is still running


jeth-ro:
### Distribution

Gentoo

### Package version

6.4.5

### Frequency

Always

### Bug description

For directories containing large numbers of files nemo can still be scanning for files after a user has initiated a copy/move which raises the file conflict dialog. If nemo has not yet scanned up to the conflicting file, the dialog is displayed partially initialized:

<img width="517" height="155" alt="Image" src="https://github.com/user-attachments/assets/8018049b-1d74-4263-bcba-c6948f311a11" />

Note the presence of both "Replace" and "Rename" options at the same time. The entry box under the expander is blank. This state persist until file scanning catches up and `file_list_ready_cb()` gets called, at which point the dialog updates to it's usual and expected state.

If the user clicks "Rename" while the dialog is in this state (perhaps expecting a further prompt) then the *empty* string in the *hidden* entry box is applied to the current directory, resolving to the parent folder. Nemo then raises a conflict dialog to replace the parent folder with the file causing the initial conflict:

<img width="441" height="408" alt="Image" src="https://github.com/user-attachments/assets/a577b6b4-5dc1-45c7-b859-3289e37865bb" />

If the user then clicks "Replace" then exactly what the dialog says will happen occurs, destroying all content in the folder in question. Yes, it the conflict dialog is pretty explicit in what is about to happen, but please note:
- this is inconsistent with the filesystem layout; the folder "Camera" does not contain a folder named "Camera", it's the same folder.
- this is inconsistent with the operation initiated by the user and their mental model of what is happening
- if things are *really* grinding slow, then the second dialog can actually look just like the first

In that last case, it's way to easy to get burned by this. For example click Rename accidentally instead of Replace, after all it's almost the same word and they both usually appear in the same place. Dialog comes looking exactly the same, and wait! No! I meant Replace! **boom** 

### Steps to reproduce

1. Create a directory containing 10000+ files on slow media (e.g. old USB disk or SD card)
2. Duplicate a few (say 10) of the files to a separate location on a different storage device
3. Copy these files to clipboard in nemo
4. Browse to the device/folder the many files
5. Paste, wait for the half empty conflict dialog
6. Select "rename" button, wait for the conflict dialog on the parent folder
7. Select "replace" button, watch everything unravel

[...]

PR #3751, opened 2026-03-31
Bugfix: rename replaces parent directory (issue #3750)

jeth-ro:
This protects against the issue described in https://github.com/linuxmint/nemo/issues/3750 by asserting that the new name used when resolving a conflict via rename cannot be blank. Also ensure that the Rename button is actually hidden prior to `file_list_ready_cb()` so as to create an initial state consistent with the show/hide logic around the replace/rename buttons in `entry_text_changed_cb()` and `checkbox_toggled_cb()`.

It kind of looks like this problem was foreseen as the `details->rename_button` button widget is hidden after creation, but that looks to be undone by the subsequent `gtk_widget_show_all()` call.

Unfortunately I wasn't actually able to test this on master as I cannot get that to run on my computer. This was fixed and tested against the v6.4.5 tag which matches my distro's current stable package. I did run `git diff master libnemo-private/nemo-file-conflict-dialog.c` while on the 6.4.5 tag and do not see anything that changes the picture. Appreciate this is not an ideal way to submit a PR, but given the changes are so trivial am hoping to get away with it...

I dare say that the whole half-baked conflict dialog while the directory scan runs is a bit odd from a UX perspective. For example, it might state in the empty space atop the dialog that the directory is still being enumerated. The expander might also be disabled until the dialog is fully valid (as it will show a blank entry, which is confusing although I think it will now at least be safe).

Am happy to further develop this PR to include such things if desired, but felt it made sense to start with a minimum fix.

### DIFF
--- a/libnemo-private/nemo-file-conflict-dialog.c
+++ b/libnemo-private/nemo-file-conflict-dialog.c
@@ -622,6 +622,7 @@ nemo_file_conflict_dialog_class_init (NemoFileConflictDialogClass *klass)
 char *
 nemo_file_conflict_dialog_get_new_name (NemoFileConflictDialog *dialog)
 {
+	g_assert(gtk_entry_get_text_length (GTK_ENTRY (dialog->details->entry)) > 0);
 	return g_strdup (gtk_entry_get_text
 			 (GTK_ENTRY (dialog->details->entry)));
 }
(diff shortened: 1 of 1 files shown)


mtwebster:
Thanks - I expanded on this a bit in 73dd03ad7f497cdd - I feel that before the dialog is fully realized you shouldn't be able to do anything but cancel (and it avoids the g_assert/termination if there was an empty entry).

PR #3759, opened 2026-04-16
build: add Docker + Makefile for dependency-free local builds

EdCordata:
Building Nemo from source requires several Cinnamon-specific dependencies (e.g., libcinnamon-desktop, libxapp) that are not consistently available across distributions, making setup more complex outside Linux Mint or in CI environments.

This PR adds three files to make building easy for anyone with Docker and
Docker Compose:

- **`Dockerfile`** — Uses `linuxmintd/mint21-amd64` as the base image.
  Since this is an official Linux Mint 21 image, all Cinnamon-specific
  packages (`libcinnamon-desktop-dev`, `libxapp-dev`, etc.) are available
  natively via apt — no third-party repo configuration needed.
- **`docker-compose.yml`** — Mounts the local source tree into the
  container and maps `./build-output/` to the meson output directory so
  compiled binaries land on the host.
- **`Makefile`** — Provides two targets:
  - `make build` — runs `meson setup` and `ninja` inside the container,
    clearing the output directory first for a clean build each time.
  - `make fix_permissions` — restores ownership of `./build-output/` and `./build/` to
    the current user, since files written by the container are owned by
    `root`.

The README is updated with a "Building from Source" section explaining how to
use this.

I’m a web developer, so this is slightly outside my area of expertise—any feedback is welcome. I added this feature for my own use, as I want to make some changes to the file manager that I personally would like to see, and I thought that this might be helpful for others as well.


mtwebster:
This doesn't really make any sense -

Sure, you can build in docker so the build dependencies are there, *but you still have matching runtime dependencies* to deal with.  Having  libxapp-dev as a build dependency means libxapp1 will be an install dependency.

cinnamon-desktop and libxapp (which isn't a Cinnamon package) are trivial to build on any Debian/Ubuntu distribution.

```bash
git clone ....
sudo mk-build-deps -ir # need equivs package installed
dpkg-buildpackage
cd ..
sudo dpkg -i *.deb
```
After you build and install those, nemo builds the same way.

EdCordata:
@mtwebster I’m not sure why this was rejected—adding Docker support would make the project buildable across all platforms and significantly simplify setup, including on Linux, and even, macOS, and Windows.

It would also reduce distro-specific build issues and the number of related pull requests.

Issue #2456, opened 2020-07-23
Nemo to show REAL filename for .desktop files


lucypol:
```
 * Nemo version (nemo --version)
Nemo 4.6.4

 * Is issue with desktop or windowed nemo?
Windowed

 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...)
Mint 20

 * 32 or 64 bit
64-bit
 ```

**Issue**
Not showing the real filename for .desktop files

**Steps to reproduce**
list files in /usr/share/applications/

**Expected behaviour**
Showing real filename

**Other information**
I would like an option to show the real filename and not this hide the real name like Windows
This is partly why we left Windows, so we can see the real information, not what is spoon fed to us by deception

Jeremy7701:
Also a problem in nemo 4.6.4 on LMDE4
How long has this .desktop thing been a "feature" of some Linux distros?
Note that you can find .desktop files scattered in multiple directories:-
`/etc/alternatives/`
`/etc/xdg/`
`/etc/xdg/autostart/`
`/home/user/.config/autostart/`
`/home/user/.kde/share/apps/`
`/home/user/.local/share/applications/`
`/usr/lib/gnome-settings-daemon-3.0/gtk-modules/`
`/usr/lib/nvidia/current/`
`/usr/share/app-install/desktop/`
`/usr/share/applications/`   etc....
**even**
`/usr/share/gnome/applications/nemo.desktop`

lucypol:
The problem is not the existence of .desktop files, it is the displaying of them, I want to see the actual filename, not the label from the contents inside it, that's windows deception shit right there.

lucypol:
> Also a problem in nemo 4.6.4 on LMDE4
> How long has this .desktop thing been a "feature" of some Linux distros?
> Note that you can find .desktop files scattered in multiple directories:-
> `/etc/alternatives/`
> `/etc/xdg/`
> `/etc/xdg/autostart/`
> `/home/user/.config/autostart/`
> `/home/user/.kde/share/apps/`
> `/home/user/.local/share/applications/`
> `/usr/lib/gnome-settings-daemon-3.0/gtk-modules/`
> `/usr/lib/nvidia/current/`
> `/usr/share/app-install/desktop/`
> `/usr/share/applications/` etc....
> **even**
> `/usr/share/gnome/applications/nemo.desktop`

You missunderstand the issue, try reading it again

Jeremy7701:
[...]

PR #3760, opened 2026-04-17
feat: add option to show real filenames

EdCordata:
This pull request adds an option to allow Nemo to display real filenames - a feature I’ve personally wanted for a long time, and one that others have requested as well:
* https://github.com/linuxmint/nemo/issues/2456
* https://forums.linuxmint.com/viewtopic.php?t=224160
* Maybe even this: https://github.com/linuxmint/nemo/issues/1404

This also solves issue of having multiple files appear with the same name, in some cases, and other issues.

The feature is **optional** and **disabled by default**, so it won’t affect existing users. It can be enabled with:
```
gsettings set org.nemo.preferences show-real-filename true
```

### DIFF
--- a/libnemo-private/nemo-file.c
+++ b/libnemo-private/nemo-file.c
@@ -230,14 +230,6 @@ nemo_file_set_display_name (NemoFile *file,
 {
 	gboolean changed;
 
-	/* When enabled, always show the actual filename instead of any custom
-	 * display name (e.g. the Name field inside .desktop files). */
-	if (custom &&
-	    file->details->name != NULL &&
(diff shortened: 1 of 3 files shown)


mmseng:
Please, please implement this. A file manager that is unable to display (nor allow editing of) a file's name is a joke.

leigh123linux:
> Please, please implement this. A file manager that is unable to display (nor allow editing of) a file's name is a joke.

Sorry we are too busy reading spammy  comments like yours, maybe if there were less it would leave us more time to do code reviews and merges!!

mmseng:
I apologize for my rude and ungrateful comment. It was made in frustration. I also recognize a need to adjust my expectations. And sorry for spamming the thread (again).

Issue #3677, opened 2026-01-10
Nemo reporting false file transfer progress for USBs


SpeeQz1:
### Distribution

Mint 22.2

### Package version

nemo 6.4.5

### Frequency

Always

### Bug description

When copying very large files likes ISOs to a USB progress may not be properly reported depending on the user's available RAM.

Usually one fix is to change the `dirty_bytes` values:

```sh
echo $((32*1024*1024)) >/proc/sys/vm/dirty_background_bytes  
echo $((1024*1024*1024)) >/proc/sys/vm/dirty_bytes
```

It would be good if `nemo` handles this visually properly.

### Steps to reproduce

1. Get a big enough USB to hold a large `.iso` file.
2. Open `nemo`.
3. Copy your `.iso` file from your computer to the USB.
4. Observe copying progress.

### Expected behavior

Nemo reports progress properly.

### Additional information

Reference discussions from here: https://askubuntu.com/questions/1117832/copying-files-from-pc-to-pendrive-gets-stuck-at-the-end-on-ubuntu-16-04

a-j-bauer:
[Written with the help of Mistral.ai]

### When copying a large file

I want to see a proper progress dialog at least every 5 seconds. Nemo is a great file manager but the basic functionality of copying or moving a file is simply **not** working right.

### Why the Progress Dialog Isn't Showing
#### Linux File Buffering Behavior

- By default, Linux uses RAM as a buffer for file operations.
- When you copy a file, the data is quickly copied into RAM (making the operation appear instantaneous to the application).
- The actual disk write happens in the background, hidden from the application.
- Since Nemo's progress tracking relies on file I/O calls (which return immediately), it sees the operation as "done" before any real disk activity occurs.

### Why Setting vm.dirty_bytes Fixes It
#### Kernel tweak

vm.dirty_bytes controls the maximum amount of dirty (modified) pages that can be held in RAM before being written to disk.
By setting it to 16MB, you force Linux to flush buffers more frequently, making the copy operation take longer and allowing Nemo's progress tracking to show actual progress.

### Why vm.dirty_bytes is a Bad Fix

Issue | Explanation
-- | --
Global Impact | Affects all file operations (e.g., dd, cp, databases, etc.).
Performance Hit | Forces more frequent disk writes, slowing down the entire system.
Not Targeted | Doesn’t solve the root issue (Nemo’s progress tracking logic).
Temporary Workaround | Only helps for files >16MB. Small files still finish too fast.

[...]

PR #3766, opened 2026-05-03
Fix: Address issue #3677

a-j-bauer:
I added an option under Edit / Preferences / Behavior  "Force synchronous file operations".
When activated, nemo uses synchronous (O_SYNC) writing and a buffer size adjustment logic to ensure the progress dialog is shown during a file write operation and that the dialog is updated frequently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,11 @@ debian/files
 *.debhelper
 
 obj-x86_64-linux-gnu
+
+# Ignore additional dev environment files 
+.vscode/
+.clang-format
+.clangd
+
+#Ignore .md files with "MY_" prefix
+MY_*.md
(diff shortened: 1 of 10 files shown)


Jeremy7701:
Do you distinguish between a data move between filesystems (which is a copy) and a move within a filesystem (which moves no data)?
Do fast devices get larger buffers than slow devices?
Do you take into account the memory requirements of background non-nemo processes (such as VM's etc)?

mtwebster:
Bring this up with the glib project maybe. The only issue lies with certain removable device types.  Applying this to *everything* is just wrong.

a-j-bauer:
@Jeremy7701:  yes I did.
@mtwebster: untrue but never mind.

a-j-bauer:
I messed up, and I’ll make sure to double-check next time. Sorry for wasting your time guys.

PR #3784, opened 2026-06-19
Fix hang when primary org.Nemo instance is frozen

loosestrife:
If service registration fails (indicating another instance owns the bus name), we check if the remote instance is responsive by sending a D-Bus Peer Ping with a 1-second timeout. If the remote instance does not respond, we log a warning and fall back to registering as G_APPLICATION_NON_UNIQUE to run locally instead of hanging indefinitely on g_application_open.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 # Meson build file
 
 # https://github.com/linuxmint/nemo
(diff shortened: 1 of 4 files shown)


mtwebster:
Hi, thanks, but in this case we need to stop the hang occurring (which is being worked on), I'm not really interested in normalizing this state.

PR #3790, opened 2026-06-25
.desktop file: use GenericName.

### LINKED ISSUE #3789: .desktop file should use GenericName for `Files` and set Name to `Nemo`.
### Distribution

Arch Linux

### Package version

6.6.4

### Frequency

Always

### Bug description

As per the specifications: https://specifications.freedesktop.org/desktop-entry/latest/recognized-keys.html

It would be preferable to set `Name` to `Nemo` and `GenericName` to `Files`.

### Steps to reproduce

Open list of applications in any DE.

### Expected behavior

See application named `Nemo` if I chose to display names and not generic names.

### Additional information

_No response_

patlefort:
As per the specifications: https://specifications.freedesktop.org/desktop-entry/latest/recognized-keys.html

Use GenericName for `Files` and set `Nemo` for Name.

Fix #3789.


clefebvre:
Files is the name we want to see for Nemo, not the app name itself. Usage of name vs generic name is not consistent enough for us to use Generic name unfortunately so I need to reject this.

This can be patched in Arch of course if Arch wants to do this.

PR #3801, opened 2026-07-05
Fix open-as-root crash

Cognomines:
# Fix crash when using **Open as Root**

## Cause

On Wayland, `nemo-desktop` uses the `open_as_admin()` code path. This function called `nemo_window_slot_open_location()` using `view->details->slot`, which ultimately resolves to `slot->pane->window`. In the desktop view context this window pointer is `NULL`, causing an assertion failure in `nemo-window-manage-views.c`.

Additionally, `open_as_admin()` constructed an invalid admin URI using `g_uri_build()`, producing `admin:/path` instead of the expected `admin:///path`. Several code paths also assumed that `nemo_view_get_uri()`, `nemo_file_get_path()`, and `g_file_get_path()` could never return `NULL`.

## Solution

This patch changes `open_as_admin()` to launch a new `nemo` process with a correctly constructed `admin://` URI, matching the behavior of the non-Wayland `open_as_root()` implementation. This avoids dereferencing the desktop view's window entirely.

It also adds `NULL` checks for URI and path retrieval functions so invalid or non-native paths are handled gracefully instead of crashing.

## Testing

Tested on Hyprland using `nemo-git` built with `meson -Dgtk_layer_shell=true`. After applying this patch, selecting **Open as Root** from both the desktop icon context menu and the desktop background context menu no longer causes `nemo-desktop` to crash.

### DIFF
--- a/src/nemo-view.c
+++ b/src/nemo-view.c
@@ -7197,19 +7197,24 @@ cb_open_as_root_watch (GPid pid, gint status, gpointer user_data)
 }
 
 static void
-open_as_admin (NemoView *view, const gchar *path) {
-    g_autoptr(GUri) uri_obj = g_uri_build (0, "admin", NULL, NULL, -1, path, NULL, NULL);
-    g_autofree gchar *uri = g_uri_to_string (uri_obj);
-    g_autoptr(GFile) location = g_file_new_for_uri (uri);
+open_as_admin (const gchar *path) {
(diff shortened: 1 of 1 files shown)


mtwebster:
I get using spawn for nemo-desktop, but there's no need for it from the windowed nemo process - the original `nemo_window_slot_open_location()` works fine.

I think we should maybe also gate this on `NEMO_IS_DESKTOP_DIRECTORY()`

Issue #2319, opened 2020-02-10
Nemo treats remote ASCII files as binary files and opens them with file-roller


mattkjames7:
```
 * Nemo version 4.4.2
 * windowed nemo
 * Distribution - Linux Mint 19.3 Cinnamon
 * NVIDIA Corporation GP104 [GeForce GTX 1070 Ti], Driver version - 435.21
 * 64 bit
 ```

**Issue**
When connected to a remote drive using SSH/SFTP (this may affect other connections too), all files which have an unknown file type, whether they are binary or ASCII, are treated as binary files and are automatically opened using file-roller.

This is quite annoying as there are many ASCII files either with an unrecognised extension, or no extension at all which would automatically be treated as text files locally, but not remotely.

**Steps to reproduce**
1. Connect to a remote server using SFTP.
2. Try to open an ASCII file which has either no extension, or an unknown extension by double-clicking it.

**Expected behaviour**
The expected behaviour I think would be to open the ASCII file with a text editor like it would to the same file stored locally.

**Other information**
A similar bug exists in nemo 4.2.2 in Linux Mint 19.2, where it prompts the user with a message saying that the file is of an unknown files type (marginally less irritating than opening file-roller).
https://github.com/linuxmint/nemo/issues/1324

nazar-pc:
For me text files with `.com` extensions are treated as binary too.

nazar-pc:
The issue is not fixed and these "stale" bots are very annoying

mtwebster:
This behavior is unfortunate, but is also deliberate by the library Nemo uses for mimetype detection - avoiding content sniffing on remote filesystems.

https://github.com/linuxmint/nemo/pull/3802#issuecomment-4893254870

nazar-pc:
Seems like a reasonable concern and wrong default behavior. Nemo could have sniffed the type on attempt to open or asked what application to use for opening every time, yet it simply defaults to doing the wrong thing instead.

PR #3802, opened 2026-07-05
Sniff remote files mimetype

leigh123linux:
This PR addresses the issue by adding a remote-file MIME sniffing fallback so that files reported as application/octet-stream over SFTP/SSH/FTP are re-evaluated from their contents (so ASCII/text files can open in a text editor instead of being treated as binary/archives).

Changes:

Introduces an async read-and-sniff path for remote (non-native) files that initially resolve to application/octet-stream.
Adds a per-file guard flag (mime_type_sniffed) to avoid repeated sniffing / prevent later overwrites back to octet-stream.
Resets the sniff guard when clearing file info.

Fixes.  #2319

### DIFF
--- a/libnemo-private/nemo-file-private.h
+++ b/libnemo-private/nemo-file-private.h
@@ -83,7 +83,9 @@ struct NemoFileDetails
 	char *symlink_name;
 	
 	GRefString *mime_type;
-	
+
+	eel_boolean_bit mime_type_sniffed : 1;
+
 	char *selinux_context;
 	char *description;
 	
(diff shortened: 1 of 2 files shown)


mtwebster:
- *if* we were to do something like this, it would belong as part of nemo-directory-async file info machinery, not as an additional async operation.
- We do not want to risk latency issues by reading file content over remote connections. This is not an imaginary concern (see https://github.com/linuxmint/nemo/commit/8454290f84904a66979fcae067e96bf6ce540d47).

Nemo's existing mimetype library calls explicitly skip content sniffing remote content, going by filename detection only. This remains the safest option.

PR #3818, opened 2026-07-30
Add gsettings option to automatically merge directories during conflicts

bones99-hub:
### What does this patch do?

This patch adds a gsettings option that automatically chooses the "Merge" action during file conflicts involving only directories. It does not have a corresponding UI element in Preferences.

### What problem does this solve?

Currently, when a copied or moved directory already exists, Nemo always prompts the user to choose how to resolve the conflict. Users who routinely merge directory trees currently need to confirm the "Merge" action every time a directory conflict occurs. This setting allows those users to opt into automatically selecting the merge action.

### Does it change existing behavior?

No. The default remains exactly the same. Only users who explicitly enable the setting see different behavior.

### Why a gsettings key?

A gsettings option was chosen so that existing behavior remains unchanged while allowing users who prefer automatic directory merging to opt in.

### Why no UI element?

This patch intentionally adds only the underlying gsettings key. Whether it should be exposed through the Preferences dialog is a separate UI decision, so this change keeps the implementation focused and leaves that discussion open.

### DIFF
--- a/libnemo-private/nemo-file-operations.c
+++ b/libnemo-private/nemo-file-operations.c
@@ -4398,6 +4398,16 @@ get_target_file_for_display_name (GFile *dir,
 	return dest;
 }
 
+static gboolean
+should_always_merge_dirs (void)
+{
+	gboolean confirm_merge;
+
+	confirm_merge = g_settings_get_boolean (nemo_preferences, NEMO_PREFERENCES_ALWAYS_MERGE_DIRS);
+
+	return confirm_merge;
+}
+
 /* Debuting files is non-NULL only for toplevel items */
 static void
 copy_move_file (CopyMoveJob *copy_job,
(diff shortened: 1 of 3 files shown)


Jeremy7701:
**Surely, it must be time for some nemo documentation of gsettings.**
`dconf` works for me as a GUI interface, but is it widely known about by new users?

mtwebster:
I feel like this is a feature for you, specifically. I can't see any reason one would want conflicts skipped automatically. That one time you wish you hadn't... it'll be too late. An extra click in a dialog isn't that intrusive to deal with.

bones99-hub:
Thanks for the feedback. I don't intend to argue the decision, but I would like to clarify one point about the behavior this patch changes.

The setting only automatically selects “Merge” when the source and destination are both directories. It does not automatically resolve file conflicts within those directories; those continue through Nemo's existing file-conflict handling. In other words, the concern about automatically overwriting a file is not part of this change.

If that's understood and the decision is simply that you don't think an opt-in preference for automatic directory merging is useful enough to maintain, that's completely fair. I'm curious, though, how come you think this is likely to be useful only to me? Directory merging is configurable in Windows' file manager as well, so I had assumed this represented a broader class of user preference than just my own workflow. I'm not trying to reopen the PR so much as understand the reasoning behind the decision.

PR #3832, opened 2026-08-31
docs: fix parameter names left from the nautilus era

darkdi:
These names date back to nautilus: `nemo_mime_activate_file` still documents `@view` and `@use_new_window`, which became `parent_window` and `flags` when the window slot work landed, and `nemo_launch_application` documents `@uris` although it takes a list of files and builds the uris itself. I left the block above `nemo_launch_application_from_command_array` alone, since it names the non-array function and moving it is a judgement call.

### DIFF
--- a/src/nemo-mime-actions.c
+++ b/src/nemo-mime-actions.c
@@ -2235,7 +2235,7 @@ activation_start_mountables (ActivateParameters *parameters)
  *
  * Activate a list of files. Each one might launch with an application or
  * with a component. This is normally called only by subclasses.
- * @view: FMDirectoryView in question.
+ * @parent_window: The window to parent dialogs on.
  * @files: A GList of NemoFiles to activate.
  *
  **/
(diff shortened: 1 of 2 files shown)


src/nemo-mime-actions.c
```diff
@@ -2235,7 +2235,7 @@ activation_start_mountables (ActivateParameters *parameters)
  *
  * Activate a list of files. Each one might launch with an application or
  * with a component. This is normally called only by subclasses.
- * @view: FMDirectoryView in question.
+ * @parent_window: The window to parent dialogs on.
  * @files: A GList of NemoFiles to activate.
```
mtwebster: This doesn't really 'fix' anything - all arguments should be listed...

parent_window
files
launch_directory
flags
user_confirmation

src/nemo-mime-actions.c
```diff
@@ -2316,9 +2316,9 @@ nemo_mime_activate_files (GtkWindow *parent_window,
  *
  * Activate a file in this view. This might involve switching the displayed
  * location for the current window, or launching an application.
- * @view: FMDirectoryView in question.
+ * @parent_window: The window to parent dialogs on.
  * @file: A NemoFile representing the file in this view to activate.
- * @use_new_window: Should this item be opened in a new window?
+ * @flags: How the location should be opened.
```
mtwebster: same

src/nemo-mime-actions.c
darkdi: Added documentation for every parameter, including slot, in signature order in eee6903. Also moved the parameter descriptions before the function description.

src/nemo-mime-actions.c
darkdi: Updated this function too in eee6903: all five parameters are documented. Verified that both parameter lists match the signatures and that executable code is unchanged.

Issue #3764, opened 2026-04-30
Nemo hangs after phone being plugged in for a while


Ames-hub:
### Distribution

Mint 22.3

### Package version

6.6.3

### Frequency

Quite often

### Bug description

Nemo hangs whenever I try to open it by the terminal or by the desktop icon.
But, the bug stopped when I unplugged my phone which Nemo had previously recognized as a storage device, despite the phone being set to "No data transfer."

### Steps to reproduce

1. Make sure Nemo is running

2. Plug in an android phone, and make sure its "Use USB for" setting is set to "No Data Transfer". Nemo will not hang right there, but it will regardless recognise the phone as a storage device, and open it to desktop to reveal an empty folder. This happens even if "No data transfer" was set even before the phone was plugged in to the computer.

3. Sorry, but this step is uncertain, not fully sure what happens to cause this bug but I suspect that it has something to do with Nemo's cache expiring or something? The bug seems to happen completely randomly, sometimes hours after an initial OS boot so I'm not sure what it could be.

4. Open Nemo. If whatever has required has happened, it'll hang. Opening Nemo in terminal will result in a timeout.

5. Unplug the phone, and Nemo will immediately open up as normal.

### Expected behavior

Instead of hanging or waiting eternally, it should recognize the drive isn't behaving properly or is refusing to connect and move on to the next drive instead of waiting an unreasonable amount of time.

### Additional information

The version of Android for the phone is Android 15.
When running "nemo" in terminal, it hangs until eventually it receives a time-out from the system.
I have been unable to find the source of the problem.
I am uncertain of whether this happens with iOS too.

mattbatt:
This is happening to me too.  I just came searching for an answer. 


> **[code]
System:
  Kernel: 6.8.0-124-generic arch: x86_64 bits: 64 compiler: gcc v: 13.3.0 clocksource: tsc
  Desktop: Cinnamon v: 6.6.7 tk: GTK v: 3.24.41 wm: Muffin v: 6.6.3 vt: 7 dm: LightDM v: 1.30.0
    Distro: Linux Mint 22.3 Zena base: Ubuntu 24.04 noble
Machine:
  Type: Desktop Mobo: Gigabyte model: B550M AORUS PRO-P serial: <superuser required>
    uuid: <superuser required> UEFI: American Megatrends LLC. v: F19 date: 10/29/2025
CPU:
[...]

PR #3833, opened 2026-09-01
nemo-places-sidebar: fetch free-space info asynchronously (#3764)

leigh123linux:
Free-space info for Home, File System, mounted volumes, and
bookmarks was fetched with a single blocking GIO call per row from
update_places() on the main thread. A gvfs mount that never answers
(e.g. an Android phone connected via MTP with "No data transfer")
froze the whole sidebar - and with it, every open Nemo window -
indefinitely.

Rows are now added with no free-space info, and an async
g_file_query_filesystem_info_async() call fills it in later if the
backend answers before the row is gone. Bookmarks pointing into a
mount are covered the same way, since they were exposed to the same
hang; the default XDG bookmarks (Pictures, Documents, etc.) are left
alone, as they're always plain subdirectories of Home rather than
distinct mounts.

mtp:// and gphoto2:// mounts are excluded from this outright rather
than queried and relied upon to error or cancel cleanly. Testing
against a real device showed that a stuck query on these backends
doesn't actually get freed by cancellation - gdb still shows a
thread parked inside the MTP backend indefinitely - so retrying or
tracking such a request is pointless and, if repeated, risks pinning
GIO's shared worker-thread pool on a backend that's never coming
back. No thread or timeout is needed for the general case: the
async call itself never blocks the caller regardless of whether or
when a well-behaved backend responds.

### DIFF
--- a/src/nemo-places-sidebar.c
+++ b/src/nemo-places-sidebar.c
@@ -150,6 +150,8 @@ typedef struct {
     guint popup_menu_action_index;
     guint update_places_on_idle_id;
 
+    GCancellable *df_cancellable;
+
 } NemoPlacesSidebar;
 
 typedef struct {
@@ -363,8 +365,8 @@ typedef struct {
        GMount *mount;
        gint index;
        gchar *tooltip;
-       gint df_percent;
-       gboolean show_df_percent;
+       GFile *df_file;
+       gboolean show_df_when_available;
 } PlaceInfo;
 
 static gint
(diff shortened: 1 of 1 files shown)


src/nemo-places-sidebar.c
```diff
@@ -442,15 +566,19 @@ add_place (NemoPlacesSidebar *sidebar,
 	   GMount *mount,
 	   int index,
 	   const char *tooltip,
-       int df_percent,
-       gboolean show_df_percent,
+       GFile *df_file,
+       gboolean show_df_when_available,
```
mtwebster, line 570: Is `show_df_when_available` even necessary? It's only ever `TRUE` if `df_file != NULL` isn't it? So we wouldn't be running `query_disk_full_async()` on a location we didn't want show it in the first place...

leigh123linux: Home is the one place they diverge. df_file is always non-NULL for Home (we always want to query its free space, for the tooltip), but show_df_when_available there is home_on_different_fs(mount_uri) — independently FALSE when Home is on the same filesystem as root:

leigh123linux: image

